PR #12, opened 2017-06-18
Add removal to frontend to support finalizers 

eborden:
Finalizers allow graphula to unwind the graph upon completion (or
failure). This allows idempotent tests to clean up after themselves. In
a more complicated database setting, this would avoid the need to
truncate many tables and repopulate any preloaded data.

### DIFF
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -140,7 +140,7 @@ runGraphulaIdempotentLogged
   -> Graph Arbitrary ToJSON nodeConstraint entity m a -> m a
 runGraphulaIdempotentLogged frontend graph = do
   graphLog <- liftIO $ newIORef ""
-  catch (go graphLog) (logFail graphLog)
+  catch (go graphLog) (logFailTemp graphLog)
     where
       go graphLog = runGraphulaIdempotent' (backendArbitraryLogged graphLog) frontend graph
 
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -217,10 +217,10 @@ finalizerFrontend finalizersRef f = case f of
   Insert n next -> do
     mEnt <- liftF $ Insert n id
     for_ mEnt $ \ent ->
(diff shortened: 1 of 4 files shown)


graphula-core/src/Graphula.hs
```diff
@@ -111,6 +115,48 @@ runGraphula frontend f =
...
+  where
+    go graphLog = runGraphulaIdempotentUsing (backendArbitraryLogged graphLog) frontend graph
+
+runGraphulaIdempotentUsing
+  :: (MonadIO m, MonadCatch m, MonadMask m)
+  => (Backend generate log (m a) -> m a)
+  -> (Frontend nodeConstraint entity (m a) -> m a)
+  -> Graph generate log nodeConstraint entity m a -> m a
+runGraphulaIdempotentUsing backend frontend f =
+  mask $ \unmasked -> do
```
eborden: I don't think this defensive masking is actually necessary.

graphula-core/src/Graphula.hs
```diff
@@ -156,14 +204,24 @@ runGraphulaReplay
...
+finalizerFrontend
+  :: (MonadThrow m, MonadIO m)
+  => IORef (FreeT (Frontend nodeConstraint entity) m ())
+  -> Frontend nodeConstraint entity (m a)
+  -> FreeT (Frontend nodeConstraint entity) m a
+finalizerFrontend finalizersRef f = case f of
+  Insert n next -> do
+    mEnt <- liftF $ Insert n id
+    for_ mEnt $ \ent ->
+      liftIO $ modifyIORef' finalizersRef ((>>) (liftF $ Remove ent ()))
```
cdparks: Why not
```haskell
liftIO $ modifyIORef' finalizerRef (liftF (Remove ent ()) >>)
```

graphula-core/src/Graphula.hs
cdparks: I don't think it's hurting anything

graphula-core/src/Graphula.hs
eborden: Yeah, that is more readable. I took it a step further and abstracted the `liftF`.

PR #13, opened 2017-06-22
Fix generic instance for nullary constructors

cdparks:
**tldr;** `ghc` couldn't find the nil `GHasDependenciesRecursive` instance for nullary constructors because of some hidden ambiguity.

#### Slightly Longer
Before, we had this weird list type:
```haskell
data List a
  = None
  | Last a
  | Cons a (List a)
```

You'll note that `None` doesn't bind `a`, so if you have an empty list, the type of `'None` is polymorphic. One way to fix this is to change `None` to `None a` and have `FindMatches` do this:
```haskell
FindMatches nodeTy depsTy () () = 'None ('NoMatch ())
```

Then, you can write this instance for `GHasDependenciesRecursive` which `ghc` can suddenly find:
```haskell
instance GHasDependenciesRecursive (Proxy ('None ('NoMatch ()))) () () where
  genericDependsOnRecursive _ () () = ()
```

But that's kind of gross.

Instead, adding a kind signature to both `FindMatches` and the null instance for `GHasDependenciesRecursive` resolves whatever ambiguity was tripping up `ghc`. In fact, we can now use the builtin type-level list. I suspect this same ambiguity is what made me add the `Last` constructor to `List` in the first place.

### DIFF
--- a/graphula-core/src/Graphula/Internal.hs
+++ b/graphula-core/src/Graphula/Internal.hs
@@ -9,20 +9,12 @@
 
 module Graphula.Internal where
 
-import Data.Proxy (Proxy(..))
-import Generics.Eot (Void)
+import Generics.Eot (Void, Proxy(..))
 import GHC.TypeLits (TypeError, ErrorMessage(..))
 
 data Match t
   = NoMatch t
   | Match t
-
--- This looks over-specified, but if we can tolerate some overlap in the type-family,
--- then we can avoid overlap in the typeclasses below.
-data List a
-  = None
-  | Last a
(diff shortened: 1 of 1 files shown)


eborden:
¡Que bueno!

Great job. I'm happy to see more built-ins and more red than green!

eborden:
Thanks @cbeav for finding this.

cbeav:
Thanks @eborden for telling me I'd found this :)

cdparks:
Tested on our API tests, which still work.

PR #15, opened 2017-06-23
Add generic HasDependencies instance for data types with no constructors

cdparks:
I'm not sure how useful this is, but the nullary constructor issue got me thinking about empty data decls. This change allows the following

```haskell
data Nope deriving Generic

instance HasDependencies Nope
```

Technically, it also allows nonsense like:

```haskell
data Nope deriving Generic

instance HasDependencies Nope where
  type Dependencies Nope = (Int, Bool)
```

but we just ignore the dependencies. There's a probably a few ways to warn the user about doing that, but the one I figured out requires overlapping instances. I doubt this is an important enough use-case to worry about that.

### DIFF
--- a/graphula-core/src/Graphula/Internal.hs
+++ b/graphula-core/src/Graphula/Internal.hs
@@ -51,9 +51,20 @@ class GHasDependenciesRecursive fieldsProxy node deps where
   genericDependsOnRecursive :: fieldsProxy -> node -> deps -> node
 
 -- This instance head only matches EoT representations of
--- datatypes with no constructors. Ignore dependencies.
-instance GHasDependencies (Proxy nodeTy) (Proxy depsTy) Void deps where
+-- datatypes with no constructors and no dependencies
+instance {-# OVERLAPPING #-} GHasDependencies (Proxy nodeTy) (Proxy depsTy) Void (Either () Void) where
   genericDependsOn _ _ node _ = node
+
+-- This instance warns the user if they give dependencies
+-- to a datatype with no constructors
+instance
+  {-# OVERLAPPABLE #-}
+  ( TypeError
+    ( 'Text "A datatype with no constructors can't use the dependencies in" ':$$:
+      DependenciesTypeInstance nodeTy depsTy
+    )
+  ) => GHasDependencies (Proxy nodeTy) (Proxy depsTy) Void (Either deps rest) where
+    genericDependsOn _ _ _ _ = error "Impossible"
 
 -- This instance head only matches EoT representations of
 -- datatypes with a single constructor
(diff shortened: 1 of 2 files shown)


graphula-core/src/Graphula/Internal.hs
```diff
@@ -51,6 +51,11 @@ class GHasDependenciesRecursive fieldsProxy node deps where
   genericDependsOnRecursive :: fieldsProxy -> node -> deps -> node
 
 -- This instance head only matches EoT representations of
+-- datatypes with no constructors. Ignore dependencies.
+instance GHasDependencies (Proxy nodeTy) (Proxy depsTy) Void deps where
```
eborden: Could this instance match on `()` and not a`deps` var?

cdparks: Here's what I was thinking. Given:
```haskell
data Nope deriving Generic
```

a user might try to set `type Dependencies Nope` to one of three things:
1. `()`, implicitly or explicitly
2. `Void`, explicitly - this isn't completely unreasonable, but then the user can't use `node @Nope`. Instead they have to say `nodeWith @Nope (undefined :: Void)`, which is probably not what they mean. We actually probably want to disallow this for all instances.
3. Anything else - this is probably a mistake. 

I figured out how to handle all three cases, but it seems to require overlapping instances:
```haskell
-- Only allow empty data decls to use () as their dependencies
instance {-# OVERLAPPING #-} GHasDependencies (Proxy nodeTy) (Proxy depsTy) Void (Either () Void) where
   genericDependsOn _ _ node _ = node

-- Disallow any other kind of non-void, non-unit dependencies for empty data decls
instance
  {-# OVERLAPPABLE #-}
  ( TypeError
    ( 'Text "A datatype with no constructors can't use the dependencies in" ':$$:
      DependenciesTypeInstance nodeTy depsTy
    )
  ) => GHasDependencies (Proxy nodeTy) (Proxy depsTy) Void (Either deps rest) where
    genericDependsOn _ _ _ _ = error "Impossible"

-- Don't let the user specify `Void` as a dependency for any node
instance
  ( TypeError
    ( 'Text "Use ‘()’ instead of ‘Void’ for datatypes with no dependencies in" ':$$:
      DependenciesTypeInstance nodeTy depsTy
    )
  ) => GHasDependencies (Proxy nodeTy) (Proxy depsTy) node Void where
    genericDependsOn _ _ _ _ = error "Impossible"
```

I'm not actually sure why the last one doesn't seem to need overlapping instances.

eborden: Hmmm, do we need to cover all cases? Or just the cases that we support?

---

This also raises the question of `Void`. I considered it when thinking about the default `Dependencies`. Here are my thoughts:

`Void` is idiomatically used when you are trying to explicitly outlaw a codepath. For example in `pipes` we use `Void` to express that something will never `yield`. So what does it mean to have `type Dependencies a = Void`?

```
x `dependsOn` y = const x $ absurd y
```

This could really be reduced to just `dependsOn = const`. The same instance for `()`.

In order to write `node` we'd need to introduce a bottom.

```
node = nodeWith $ error "can never have dependencies"
```

So, it might be "morally" correct to use `Void`. in a sense it is absurd for something with no dependencies to depend on anything. We should never being inspecting non-existent dependencies.

`()` on the other hand plays nicely into the algebra of types. `()` is the identity object of product types. It erases itself by carrying no semantic meaning. However, there really is nothing in graphula forcing us to produce just product types and then `()` is inappropriate because it does carry meaning in association with sum types.

---

`type Dependencies a = Void` also has a nice property that it forces `nodeWith` to be impossible to utilize without a bottom, so users wouldn't accidentally litter `nodeWith ()` around their codebase.

eborden: Talked offline. We decided that the improved errors trump the mild increase in complexity of these overlapping instances. :shipit:

Issue #7, opened 2017-04-19
Dump JSON as an array


eborden:
JSON data is currently dumped as an object per line. We could make this much simpler by dumping an array of objects. This also allows room for pretty printing making it more human readable.

PR #16, opened 2017-06-24
Store logged nodes in a Seq and dump it to an array of JSON objects

cdparks:
`test.graphula` before:
```plaintext
{"ab":5,"aa":"success"}
```

`test.graphula` after:
```plaintext
[{"ab":5,"aa":"success"}]
```

Per issue #7


eborden:
LGMT. Thanks for addressing this.

PR #19, opened 2017-11-07
 Add a type family for defining constraints over multiple nodes

eborden:
This makes utility functions a lot more compact by allowing nodes to be
defined within a list instead of `(NodeGraphula n, ...NodeGraphula n+1)`.

I'm not 100% sold on the name of this constraint since `GraphulaT`
already exists as a `Transformer` and `Graphula` would infer the
non-transformer version. 

`GraphulaContext` seems naturally replaceable since it is essentially 
`Graphula m '[a]`.

```
type GraphulaContext m a = (MonadGraphula m, GraphulaNode m a)
```

Though removing `MonadGraphula` and replacing it with this is an option
since all operations over `MonadGraphula` contain a `NodeConstraint`
as well.

The only function that uses `MonadGraphula`, and not `GraphulaContext` is:

```
attemptsToInsertWith
  :: forall a m . (NodeConstraint m a, Typeable a, MonadGraphula m)
  => Int
  -> m a
  -> m (Node m a)
```

This points to `MonadGraphula` and `NodeConstraint` being coupled and The
distinction between `GraphulaContext` and `MonadGraphula` being useless.


graphula-core/src/Graphula.hs
```diff
@@ -69,28 +72,28 @@ module Graphula
   , GenerationFailure(..)
   ) where
 
-import Prelude hiding (readFile, lines)
```
pbrisbin: :lipstick: Keep `Prelude` at the top in its own group.

graphula-core/src/Graphula.hs
```diff
@@ -101,6 +104,21 @@ type MonadGraphula m =
   , MonadThrow m
   )
 
+-- | A constraint over lists of nodes for 'MonadGraphula', and 'GraphulaNode'.
+--   Helpful for defining utility functions over many nodes.
```
pbrisbin: FWIW: I'd format this as:

```hs
-- | A constraint over lists of nodes for @'MonadGraphula'@ and @'GraphulaNode'@
--
-- Helpful for defining utility functions over many nodes.
--
-- @
-- ...
-- @
--
type family ...
```

Which follow these guides:

- Have a short first line, not punctuated (like a commit message)
- Keep a line of whitespace before the "body" (like a commit message), which makes it not-annoying to
- Align the "body" normally
- For docs with bodies, put another line of "whitespace" before the function (I think it's also generally advised to surround code blocks with "whitespace", which applies here as well)
- Use proper "markdown" (the haddock kind)

cdparks:
Neat. What happens to inference if you take the type signature off of declaration containing a bunch of `node[Edit][With]`s? Do we end up with separate `GraphulaContext m [a]`'s or does it normalize `GraphulaContext`?

eborden:
This specific function fails to compile because of `Identity`

```haskell
mkABC :: (GraphulaContext m '[A, B, C], Node m ~ Identity) => m (Node m C)
mkABC = do
  Identity a <- node @A
  Identity b <- nodeWith @B (only a)
  nodeEditWith @C (a, b) $ \n ->
    n { cc = "spanish" }
```

```haskell
mkABC = do
  Identity a <- node @A
  Identity b <- nodeWith @B (only a)
  nodeEditWith @C (a, b) $ \n ->
    n { cc = "spanish" }
```

```
   /home/eborden/code/graphula/graphula-core/README.lhs:180:17: error:
        • Couldn't match type ‘Node m0’ with ‘Identity’
          Expected type: m0 (Identity A)
            Actual type: m0 (Node m0 A)
          The type variable ‘m0’ is ambiguous
        • In a stmt of a 'do' block: Identity a <- node @A
          In the expression:
            do Identity a <- node @A
               Identity b <- nodeWith @B (only a)
               nodeEditWith @C (a, b) $ \ n -> n {cc = "spanish"}
          In an equation for ‘mkABC’:
              mkABC
                = do Identity a <- node @A
                     Identity b <- nodeWith @B (only a)
                     nodeEditWith @C (a, b) $ \ n -> ...
        • Relevant bindings include
            mkABC :: m0 (Node m0 C) (bound at README.lhs:179:1)
        |
    180 |   Identity a <- node @A
        |                 ^^^^^^^
```

eborden:
Though if I drop it into a concrete context it gets this signature:

```
mkABC :: GraphulaT GraphulaIdentity C
```

eborden:
Turn of the monomorphism restriction and we get:
```
      mkABC :: (Node m ~ Identity, MonadThrow m, MonadGraphulaFrontend m,
                MonadGraphulaBackend m, NodeConstraint m C, Logging m C,
                Generate m C, NodeConstraint m B, Logging m B, Generate m B,
                NodeConstraint m A, Logging m A, Generate m A) =>
               m C
```

cdparks:
Ok, that's fine. Was just curious what it looked like.

eborden:
Shall I merge?

PR #20, opened 2018-07-19
Update to lts-11.17

pbrisbin:
Notable changes:

- Updates for persistent-2.8
- MonadUnliftIO instead of Catch/Throw/Mask/BaseControl


graphula-core/bench/Main.hs
```diff
@@ -1,19 +1,19 @@
-{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 module Main where
 
+import Control.Monad (replicateM, void)
+import Control.Monad.Catch
```
pbrisbin: Not sure why this is still here. Will investigate.

graphula-core/src/Graphula.hs
```diff
@@ -71,14 +70,23 @@ module Graphula
   , GenerationFailure(..)
   ) where
 
-import Prelude hiding (lines, readFile)
+import Prelude hiding (readFile)
 
-import Control.Exception (Exception, SomeException)
-import Control.Monad.Catch (MonadCatch(..), MonadMask(..), MonadThrow(..), bracket)
+--import Control.Exception (Exception, SomeException)
```
pbrisbin: Will remove.

graphula-core/src/Graphula.hs
```diff
@@ -167,7 +178,15 @@ instance (Monad m, MonadGraphulaFrontend m) => MonadGraphulaFrontend (GraphulaT
...
-  deriving (Functor, Applicative, Monad, MonadThrow, MonadCatch, MonadMask, MonadIO, MonadReader (IORef (m ())))
+  deriving (Functor, Applicative, Monad, MonadIO, MonadReader (IORef (m ())))
+
+instance MonadUnliftIO m => MonadUnliftIO (GraphulaIdempotentT m) where
+  {-# INLINE askUnliftIO #-}
+  askUnliftIO = GraphulaIdempotentT $ withUnliftIO $ \u ->
+    return $ UnliftIO $ unliftIO u . runGraphulaIdempotentT'
+  {-# INLINE withRunInIO #-}
+  withRunInIO inner = GraphulaIdempotentT $ withRunInIO $ \run ->
+    inner $ run . runGraphulaIdempotentT'
```
pbrisbin: FWIW: I just copied the `ReaderT` implementation and followed the type errors. I totally don't understand how all this works.

graphula-core/src/Graphula.hs
```diff
@@ -445,7 +464,7 @@ attemptsToInsertWith attempts source
 
 -- | For entities that only have singular 'Dependencies'. It uses data instead
 -- of newtype to match laziness of builtin tuples.
-data Only a = Only { fromOnly :: a }
+newtype Only a = Only { fromOnly :: a }
```
cdparks: @pbrisbin Comment is out of date now - this was using `data` on purpose to match builtin tuples, but it probably has limited utility

graphula-core/src/Graphula.hs
eborden: Yeah, this is essentially a `newtype` deriving of this instance. Could you just `newtype` derive this?

graphula-persistent/src/Graphula/Persist.hs
```diff
@@ -1,33 +1,42 @@
...
-  deriving (Functor, Applicative, Monad, MonadIO, MonadThrow, MonadCatch, MonadMask, MonadReader (RunDB backend n m))
+  deriving (Functor, Applicative, Monad, MonadIO, MonadReader (RunDB backend n m))
+
+instance MonadUnliftIO m => MonadUnliftIO (GraphulaPersistT backend n m) where
+  {-# INLINE askUnliftIO #-}
+  askUnliftIO = GraphulaPersistT $ withUnliftIO $ \u ->
+    return $ UnliftIO $ unliftIO u . runGraphulaPersistT'
+  {-# INLINE withRunInIO #-}
+  withRunInIO inner = GraphulaPersistT $ withRunInIO $ \run ->
+    inner $ run . runGraphulaPersistT'
```
eborden: This should also just be `newtype` deriving.

graphula-persistent/src/Graphula/Persist.hs
```diff
@@ -1,33 +1,42 @@
...
 
 instance MonadTrans (GraphulaPersistT backend n) where
   lift = GraphulaPersistT . lift
 
-instance (MonadIO m, Applicative n, MonadIO n, SqlBackendCanWrite backend) => MonadGraphulaFrontend (GraphulaPersistT backend n m) where
+instance (MonadIO m, Applicative n, MonadIO n, SqlBackendCanWrite backend, BaseBackend backend ~ SqlBackend) => MonadGraphulaFrontend (GraphulaPersistT backend n m) where
```
eborden: Should this be the compact class instead of `BaseBackend`?

graphula-core/src/Graphula.hs
pbrisbin: No, GHC spat an incomprehensible error when I tried that.

graphula-persistent/src/Graphula/Persist.hs
pbrisbin: Can you unpack "the compact class"?

My understanding of this is that persistent is broken such that some functions need `SqlBackendCanWrite` (`BackendCompatible`), but other functions still need `BaseBackend ~ SqlBackend`, so we need both constraints even though they're clearly redundant. It's possible now that we've moved to `SqlPersistT` uniformly, there's no use for `SqlBackendCanWrite` -- the feedback loop in trying out the different variations is a bit slow.

5outh:
<3

PR #21, opened 2018-11-30
Rough draft of monommorphing to persistent

eborden:
Supporting sequential and non-sequential keys adds a level of complexity
that is not bearable. Instead of remaining abstract, we can just
monomorph this tool to persistent and throw out a bunch of complexity.


graphula-core/src/Graphula.hs
```diff
@@ -131,16 +130,12 @@ type family GraphulaContext (m :: Type -> Type) (ts :: [Type]) :: Constraint whe
...
-  -- ^ A constraint applied to nodes. This is utilized during
-  --   insertion and can be leveraged by frontends with typeclass interfaces
-  --   to insertion.
-  type Node m :: * -> *
-  -- ^ A wrapper type used to return relevant information about a given
-  --   node. `graphula-persistent` returns all nodes in the 'Database.Persist.Entity' type.
-  insert :: NodeConstraint m a => a -> m (Maybe (Node m a))
-  remove :: NodeConstraint m a => Node m a -> m ()
+class MonadIO m => GraphulaRunDB m where
+  runDB :: ReaderT SqlBackend m a -> m a
```
cdparks: Should we steal this fairly common identifier?

graphula-core/src/Graphula.hs
```diff
@@ -190,16 +179,16 @@ instance MonadUnliftIO m => MonadUnliftIO (GraphulaIdempotentT m) where
 instance MonadTrans GraphulaIdempotentT where
   lift = GraphulaIdempotentT . lift
 
-instance (Monad m, MonadIO m, MonadGraphulaFrontend m) => MonadGraphulaFrontend (GraphulaIdempotentT m) where
-  type NodeConstraint (GraphulaIdempotentT m) = NodeConstraint m
-  type Node (GraphulaIdempotentT m) = Node m
+  {-
+  TODO figure out how to support tracking keys.
```
eborden: Not having `MonadGraphulaFrontend` breaks our ability to handle idempotent graphs, since we cannot overload `insert` based on the the context.

Possibly we want to key `MonadGraphulaFrotend` and redefine `insert` as
```
insert :: Maybe (Key a) -> a -> m (Entity a)
```

cdparks: Actually with everything else monomorphed, I think you can keep the frontend class and recover idempotent graphs

```haskell
class MonadGraphulaFrontend m where
  type NodeConstraint m a :: * -> Constraint
  -- not sure about this constraint might need to have constraints on the Key as well
  insert :: NodeConstraint m a => Maybe (Key a) -> a -> m (Maybe (Entity a))
  remove :: NodeConstraint m a => Key a -> m ()
```

graphula-core/src/Graphula.hs
```diff
@@ -430,34 +410,38 @@ nodeEditWith dependencies edits =
...
+  -> m (Entity a)
 attemptsToInsertWith attempts source
   | 0 >= attempts =
       throwIO . GenerationFailureMaxAttempts $ typeRep (Proxy :: Proxy a)
   | otherwise = do
     value <- source
-    insert value >>= \case
+    mKey <- liftIO $ generate genEntityKey
+    mNode <- runDB $ getEntity =<< case mKey of
+      Nothing -> insert value
```
cdparks: Recall that we need to check unique constraints here. `insert` will throw if unique constraints are violated.

graphula-core/src/Graphula.hs
cdparks: Haha jinx

eborden:
Just gotta bring back `README.md` and I'll be ready to merge this.

PR #23, opened 2019-02-16
Add constraint based editing

eborden:
Based on https://github.com/frontrowed/graphula/issues/22

`suchThat` is a powerful tool in property based testing. Allowing for
patterns like `suchThat` enables more powerful testing with `graphula`.
This is a proof of concept that proposes changing `edit` to a combinator
of type `a -> Maybe a` instead of `a -> a`. With the addition of `Maybe`
we can allow for failure in an edit, which leads to constraint based
combinators like `suchThat`.

```hs
edit :: (a -> a) -> a -> Maybe a
edit f = Just . f

suchThat :: (a -> Bool) -> a -> Maybe a
suchThat f x = x <$ guard (f x)
```

### DIFF
--- a/graphula-core/README.lhs
+++ b/graphula-core/README.lhs
@@ -122,11 +122,11 @@ loggingAndReplaySpec = do
     logFile = "test.graphula"
     -- We'd typically use `runGraphulaLogged` which utilizes a temp file.
     failingGraph = runGraphulaT runDB . runGraphulaLoggedWithFileT logFile $ do
-      Entity _ a <- nodeEdit @A $ \n ->
+      Entity _ a <- node @A . edit $ \n ->
         n {aA = "success"}
       liftIO $ aA a `shouldBe` "failed"
     replayGraph = runGraphulaT runDB . runGraphulaReplayT logFile $ do
-      Entity _ a <- node @A
+      Entity _ a <- node @A Just
       liftIO $ aA a `shouldBe` "success"
 
   failingGraph
@@ -141,12 +141,12 @@ simpleSpec :: IO ()
 simpleSpec =
   runGraphulaT runDB $ do
     -- Declare the graph at the term level
-    Entity aId _ <- node @A
+    Entity aId _ <- node @A Just
     liftIO $ putStrLn "A"
-    Entity bId b <- nodeWith @B (only aId)
+    Entity bId b <- nodeWith @B (only aId) Just
     -- Type application is not necessary, but recommended for clarity.
     liftIO $ putStrLn "B"
-    Entity _ c <- nodeEditWith @C (aId, bId) $ \n ->
+    Entity _ c <- nodeWith @C (aId, bId) . edit $ \n ->
       n { cC = "spanish" }
     liftIO $ putStrLn "C"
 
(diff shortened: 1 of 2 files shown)


eborden:
This could also be plugged directly into `Arbitrary` by changing `generate`'s signature.

```diff
class MonadGraphulaBackend m where
  type Logging m :: * -> Constraint
  -- ^ A constraint provided to log details of the graph to some form of
  --   persistence. This is used by 'runGraphulaLogged' to store graph nodes as
  --   JSON 'Value's.
  type Generate m :: * -> Constraint
  -- ^ A constraint for pluggable node generation. 'runGraphula'
  --   utilizes 'Arbitrary', 'runGraphulaReplay' utilizes 'FromJSON'.
--  generateNode :: Generate m a => m a
++  generateNode :: Generate m a => (a -> Maybe a) -> m a
  logNode :: Logging m a => a -> m ()

instance MonadIO m => MonadGraphulaBackend (GraphulaT n m) where
  type Logging (GraphulaT n m) = NoConstraint
  type Generate (GraphulaT n m) = Arbitrary
--  generateNode = liftIO $ generate arbitrary
++  generateNode f = liftIO . generate $ (f<$> arbitrary) `suchThat` isJust
  logNode _ = pure ()

```

eborden:
@cdparks do you have ideas for function names?

We had:
```
node
nodeWith
nodeEdit
nodeEditWith
```

Is there a way to get the same 4 combinators that doesn't include the misnomer `edit`?

cdparks:
Some nasty type classes with overlapping instances? :grimacing:

eborden:
I mean if you are going to do that then you might as well use type families to get var args, which I also wouldn't be very psyched about. I'd rather just give these things some expected names rather then complex types.

5outh:
I agree with this:

> I think this looks ok, but I don't love how nodes that don't need any edits have to pay for this with an extraneous `Just`.

Are we against adding combinators that default the `a -> Maybe a` argument? The API for graphula is still super thin; I think if this doesn't exist as a part of the library, I'd just add it to `FrontRow.Graphula` anyway under some random name.

Right now we have the following:

```haskell
node -- defaults `Dependencies` to `()`, requires "edit"
nodeWith -- specify `Dependencies`, requires "edit"
```

Notably the important bits are the dependencies and the "edit" function (whatever you want to call it). If we want to constrain the ways a node can be created, we can accept both of these under a data type in `With`? e.g.

```haskell
data NodeOptions a = NodeOptions
  { dependencies :: First (Dependencies a)
  , constraint :: a -> Maybe a
  }

instance Monoid (NodeOptions a)

depend :: Dependencies a -> NodeOptions a
constrain :: (a -> Maybe a) -> NodeOptions a

node = nodeWith $ depend ()

-- then we get something like
nodeWith @Foo $ depend (barId, bazId)
nodeWith @Bar $ depend (fooId, bazId) <> constrain (suchThat isBigBar)
```

I'm just spitballing here, but I'm just trying to think about what makes `Graphula` tick. `With` is a general term, but we are using it very specifically to mean "providing dependencies." Maybe it can be generalized to accept both pieces of important information in this way. WDYT?

eborden:
Hmm, `NodeOptions` is an interesting avenue. I have to roll this around in my head more.

eborden:
Closing this as @cdparks' PR merged enabling constraint based editing.

PR #25, opened 2019-04-15
Add example to run the lhs file

adomokos:
I added a short note to the bottom of the lhs file describing how the
user could run the examples in this file. This could have helped me, as
I was unfamiliar with the `markdown-unlit` package.

### DIFF
--- a/graphula-core/README.lhs
+++ b/graphula-core/README.lhs
@@ -193,3 +193,5 @@ runDB f = runSqlite "test.db" $ do
   f
 ```
 -->
+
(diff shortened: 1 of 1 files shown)


adomokos:
I [see](https://github.com/freckle/graphula/blob/master/graphula-core/package.yaml#L40-L53). I was looking for a `test` directory, never checked `ide targets` or looked at `package.yaml`.

Yes, I think mentioning `stack test` would be helpful. I'll update the file with that content.

adomokos:
Not sure if we need it. I am closing it for now. Thanks!

PR #26, opened 2019-05-02
Fix law statement

eborden:
The law for `HasDependencies` was not well formed. This fixes it.


eborden:
Sadly I don't think anything does, but it was the original intention of the `prop>` syntax.

PR #27, opened 2019-11-01
Ensure embedded keys are used to check uniqueness

cdparks:
When a node implements `EntityKeyGen` such that `genEntityKey` returns `Just`, we use `insertKey` to write the node to the database. However, if the key is a composite key, `persistent` keeps those values in the record itself, and `insertKey` doesn't ensure that the key argument and the record agree on those values.

This is bad.

Instead of trying to bake this into the API, we just do the right thing at runtime. We do the following:

1. Generate a node
2. Optionally generate a key
3. Check - did we generate a key, and does this type have a `Primary` definition?
    1. No! We're done! Insert away!
    2. Yes!
        1. Convert the record to an association list of `[(HaskellName, PersistValue)]`
        2. Convert the key to an association list of `[(HaskellName, PersistValue)]`
        3. Replace any values in the record list with values from the key list where the names match up
        4. Insert!

This is a little wacky, but it's basically what `persistent` does on write and read anyway. We'll only pay for it with entities that implement `genEntityKey` and declare `Primary`, so I don't think it's much of a problem.

I did go ahead and merge `HasDependencies` and `EntityKeyGen`.

### DIFF
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -25,6 +25,7 @@
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE KindSignatures #-}
@@ -33,8 +34,10 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 module Graphula
   ( -- * Graph Declaration
@@ -55,7 +58,7 @@ module Graphula
   , MonadGraphulaBackend(..)
   , MonadGraphulaFrontend(..)
   , EntityKeyGen(..)
-  , emplaceCompositeKey
+  , KeyTag(..)
   -- ** Backends
   , runGraphulaT
   , GraphulaT
@@ -139,6 +142,9 @@ type family GraphulaContext (m :: Type -> Type) (ts :: [Type]) :: Constraint whe
    GraphulaContext m (t ': ts) = (GraphulaNode m t, GraphulaContext m ts)
 
 class EntityKeyGen a where
+  -- | Either @'SimpleKey'@ or @'CompositeKey'@
+  type KeyType a :: KeyTag
+
   -- | Explicitly generate a key for a node
   genEntityKey :: Gen (Maybe (Key a))
   genEntityKey = pure Nothing
@@ -150,30 +156,8 @@ class EntityKeyGen a where
   -- key and the columns in the node are actually the same.
   --
   emplaceKey :: a -> Key a -> a
-  emplaceKey a _ = a
-
--- | Update the node with its generated composite key
---
--- We can't this in general because non-composite keys won't have a
--- @'Generic'@ instance because @'SqlBackendKey'@ doesn't have a
--- @'Generic'@ instance. Therefore, we place it outside the
--- @'EntityKeyGen'@ class and allow clients to specify it when
--- necessary.
---
-emplaceCompositeKey
-  :: forall a
-   . ( HasEot a
-     , HasEot (Key a)
-     , GHasDependencies (Proxy a) (Proxy (Key a)) (Eot a) (Eot (Key a))
-     )
-  => a
-  -> Key a
-  -> a
-emplaceCompositeKey a key = fromEot $ genericDependsOn
-  (Proxy :: Proxy a)
-  (Proxy :: Proxy (Key a))
-  (toEot a)
-  (toEot key)
+  default emplaceKey :: EmplaceKeyTagged (KeyType a) a => a -> Key a -> a
+  emplaceKey = emplaceKeyTagged (Proxy @(KeyType a))
 
 class MonadGraphulaBackend m where
   type Logging m :: * -> Constraint
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -55,6 +55,7 @@ module Graphula
   , MonadGraphulaBackend(..)
   , MonadGraphulaFrontend(..)
   , EntityKeyGen(..)
+  , emplaceCompositeKey
   -- ** Backends
   , runGraphulaT
   , GraphulaT
@@ -138,8 +139,41 @@ type family GraphulaContext (m :: Type -> Type) (ts :: [Type]) :: Constraint whe
    GraphulaContext m (t ': ts) = (GraphulaNode m t, GraphulaContext m ts)
 
 class EntityKeyGen a where
+  -- | Explicitly generate a key for a node
   genEntityKey :: Gen (Maybe (Key a))
   genEntityKey = pure Nothing
+
+  -- | Update the node with its generated key if applicable
+  --
+  -- This is important when generating nodes that have a generated
+  -- compound key since persistent doesn't check that the compound
+  -- key and the columns in the node are actually the same.
+  --
+  emplaceKey :: a -> Key a -> a
+  emplaceKey a _ = a
+
+-- | Update the node with its generated composite key
+--
+-- We can't this in general because non-composite keys won't have a
+-- @'Generic'@ instance because @'SqlBackendKey'@ doesn't have a
+-- @'Generic'@ instance. Therefore, we place it outside the
+-- @'EntityKeyGen'@ class and allow clients to specify it when
+-- necessary.
+--
+emplaceCompositeKey
+  :: forall a
+   . ( HasEot a
+     , HasEot (Key a)
+     , GHasDependencies (Proxy a) (Proxy (Key a)) (Eot a) (Eot (Key a))
+     )
+  => a
+  -> Key a
+  -> a
+emplaceCompositeKey a key = fromEot $ genericDependsOn
+  (Proxy :: Proxy a)
+  (Proxy :: Proxy (Key a))
+  (toEot a)
+  (toEot key)
 
 class MonadGraphulaBackend m where
   type Logging m :: * -> Constraint
@@ -505,7 +539,8 @@ attemptsToInsertWith attempts source
   | otherwise = do
     value <- source
     mKey <- liftIO $ generate genEntityKey
-    insert mKey value >>= \case
+    let updated = maybe value (emplaceKey value) mKey
+    insert mKey updated >>= \case
       Just a -> pure a
       Nothing -> pred attempts `attemptsToInsertWith` source
 
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -25,7 +25,6 @@
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE KindSignatures #-}
@@ -34,10 +33,8 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
-{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
-{-# LANGUAGE UndecidableInstances #-}
 
 module Graphula
   ( -- * Graph Declaration
@@ -57,7 +54,6 @@ module Graphula
   , MonadGraphula
   , MonadGraphulaBackend(..)
   , MonadGraphulaFrontend(..)
-  , KeyTag(..)
   -- ** Backends
   , runGraphulaT
   , GraphulaT
@@ -85,24 +81,37 @@ import Control.Monad.Trans (MonadTrans, lift)
 import Data.Aeson (FromJSON, Result(..), ToJSON, Value, eitherDecodeStrict', encode, fromJSON, toJSON)
 import Data.ByteString (readFile)
 import Data.ByteString.Lazy (hPutStr)
+import Data.Either (fromRight)
 import Data.Foldable (for_)
 import Data.IORef (IORef, modifyIORef', newIORef, readIORef, writeIORef)
 import Data.Kind (Type)
+import Data.Maybe (fromMaybe, isNothing)
 import Data.Proxy (Proxy(..))
 import Data.Sequence (Seq, ViewL(..), empty, viewl, (|>))
 import Data.Typeable (TypeRep, Typeable, typeRep)
 import Database.Persist
   ( Entity
+  , HaskellName
   , Key
   , PersistEntity
   , PersistEntityBackend
+  , PersistValue
   , checkUnique
   , delete
+  , entityDef
+  , entityFields
   , entityKey
+  , entityKeyFields
+  , entityPrimary
+  , fieldHaskell
+  , fromPersistValues
   , get
   , getEntity
   , insertKey
   , insertUnique
+  , keyToValues
+  , toPersistFields
+  , toPersistValue
   )
 import Database.Persist.Sql (SqlBackend)
 import Generics.Eot (Eot, HasEot, fromEot, toEot)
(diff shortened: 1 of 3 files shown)


cdparks:
The other alternative would be to get `persistent` to implement `insertKey` differently

cdparks:
Mmm, even if we made `insertKey` behave, we'd have to change how we check uniques ahead of time:

```haskell
whenNothing existingKey $ do
  existingUnique <- checkUnique n -- <-------- garbage in n
  whenNothing existingUnique $ do
    insertKey key n -- <-------- silently fixes n to use values from key too late
    getEntity key
```

graphula-core/src/Graphula.hs
```diff
@@ -138,9 +139,42 @@ type family GraphulaContext (m :: Type -> Type) (ts :: [Type]) :: Constraint whe
...
   genEntityKey = pure Nothing
 
+  -- | Update the node with its generated key if applicable
+  --
+  -- This is important when generating nodes that have a generated
+  -- compound key since persistent doesn't check that the compound
+  -- key and the columns in the node are actually the same.
+  --
+  emplaceKey :: a -> Key a -> a
+  emplaceKey a _ = a
```
eborden: The current trend in this typeclass is to provide defaults for the 90% case. I think I support the default here.

graphula-core/src/Graphula.hs
```diff
@@ -138,9 +139,42 @@ type family GraphulaContext (m :: Type -> Type) (ts :: [Type]) :: Constraint whe
...
   genEntityKey :: Gen (Maybe (Key a))
   genEntityKey = pure Nothing
 
+  -- | Update the node with its generated key if applicable
+  --
+  -- This is important when generating nodes that have a generated
+  -- compound key since persistent doesn't check that the compound
+  -- key and the columns in the node are actually the same.
+  --
+  emplaceKey :: a -> Key a -> a
```
eborden: Purely aesthetic. I'm not crazy about the name `emplace`, I also think `embed` might be more correct.

graphula-core/src/Graphula.hs
```diff
@@ -138,9 +139,42 @@ type family GraphulaContext (m :: Type -> Type) (ts :: [Type]) :: Constraint whe
    GraphulaContext m (t ': ts) = (GraphulaNode m t, GraphulaContext m ts)
 
 class EntityKeyGen a where
```
eborden: Big mood, should `EntityKeyGen` and `HasDependencies` be collapsed into the same typeclass. They are both required for any entity, so keeping them separate is just more boilerplate.

eborden: This isn't necessary work for this PR.

cdparks:
I changed my mind about how to do this. New description incoming.

graphula-core/src/Graphula.hs
```diff
@@ -193,14 +201,36 @@ instance (MonadIO m, Applicative n, MonadIO n) => MonadGraphulaFrontend (Graphul
         Just key -> do
           existingKey <- get key
           whenNothing existingKey $ do
-            existingUnique <- checkUnique n
+            let updated = embedKey key n
```
cdparks: We could also do this out in `attemptsToInsertWith`. I think in either case, we'll end up logging the unchanged node, so we might wanna rejigger when we log.

graphula-core/src/Graphula.hs
```diff
@@ -193,14 +201,36 @@ instance (MonadIO m, Applicative n, MonadIO n) => MonadGraphulaFrontend (Graphul
...
+    (fieldHaskell <$> entityFields def)
+    (toPersistValue <$> toPersistFields n)
+ where
+  def = entityDef $ Just n
+
+  keyMap :: [(HaskellName, PersistValue)]
+  keyMap = zip (fieldHaskell <$> entityKeyFields def) (keyToValues key)
+
+  replaceField :: HaskellName -> PersistValue -> PersistValue
+  replaceField name value = fromMaybe value $ lookup name keyMap
```
cdparks: Most of this is modified from [`entityValues` in `persistent`](https://www.stackage.org/haddock/lts-14.12/persistent-2.9.2/Database-Persist-Class.html#v:entityValues).

cdparks:
Hmm, do we like that restyled is using `ormolu`? Or should I copy the config from `megarepo` into here?

eborden:
We should turn off `ormolu`.

graphula-core/src/Graphula.hs
```diff
@@ -193,14 +201,36 @@ instance (MonadIO m, Applicative n, MonadIO n) => MonadGraphulaFrontend (Graphul
...
     RunDB runDB <- ask
     lift . runDB $ delete key
 
+-- | Embed generated key components in record
+--
+-- @'insertKey'@ doesn't do this for us, and we want to update the
+-- record before doing a @'checkUnique'@ anyway.
+--
+embedKey :: PersistEntity a => Key a -> a -> a
+embedKey key n = if isNothing $ entityPrimary def
```
eborden: How about `record` instead of `n`?

graphula-core/src/Graphula.hs
```diff
@@ -193,14 +201,36 @@ instance (MonadIO m, Applicative n, MonadIO n) => MonadGraphulaFrontend (Graphul
...
+  then n
+  else fromRight n $ fromPersistValues $ zipWith
+    replaceField
+    (fieldHaskell <$> entityFields def)
+    (toPersistValue <$> toPersistFields n)
+ where
+  def = entityDef $ Just n
```
eborden: Is there a clearer name than `def`? Is this a `definition`? `def` generally means `default` in Haskell land.

pbrisbin:
It's unintentional that `ormolu` ran by default (not here, like in Restyled at all). I'll fix that.

graphula-core/src/Graphula.hs
cdparks: Maybe `entDef`? These are the distinct names I see used in `persistent`:

```plaintext
let t = entityDef $ dummyFromUnique u
let entDef = entityDef $ Just record
let entity = entityDef $ Just record
let entityDef' = entityDef records
let ent = entityDef $ Just record
```

cdparks:
Re: upstreaming, it only makes sense as a part of `insertKey`, but I suspect this use-case isn't very common. In "real" applications, it's unlikely that all of these would be true:
1. You have a real record
2. You have a real key
3. `persistent` embeds a key in the record
4. The key you have does not match the key embedded in the record

This is entirely due to us separating record and key generation.

Additionally, doing this in `insertKey` is actually too late for us since we need to `checkUnique` ahead of time:

```haskell
whenNothing existingKey $ do
  let updated = embedKey key n
  existingUnique <- checkUnique updated -- must check unique constraints here
  whenNothing existingUnique $ do
    insertKey key updated -- too late to embed key
    getEntity key
```

I did think maybe a better API would be to make `genEntityKey` depend on the value so you can just pull it out of the record if it's generated as part of the `Arbitrary` instance.

```haskell
class HasDependencies a where
  type Dependencies a :: *

  ...

  genEntityKey :: a -> Gen (Maybe (Key a))
```

Unfortunately, we still need to call `embedKey` after this in the cases where you implement this, but it doesn't actually depend on the record. Alternatively, we define this:

```haskell
class HasDependencies a where
  type Dependencies a :: *

  ...

  extractKey :: a -> Maybe (Key a) -- Pure, never need to call embedKey
  extractKey _ = Nothing

  genKey :: Gen (Maybe (Key a)) -- Has access to Arbitrary, must call embedKey
  genKey = pure Nothing
```

Then we do this:

```haskell
attemptsToInsertWith attempts source =
  ...
    value <- source
    mKey <- liftIO $ maybe ((value `embedKey`) <$> generate genKey) pure $ extractKey value
    insert mKey value >>= \case
      Just a -> pure a
      Nothing -> pred attempts `attemptsToInsertWith` source
```

And we only pay for `embedKey` when the user defines `genKey` (which would be rare now).

WYT?

eborden:
There are some important effects of embeding vs coalescing a key. In the case where you have a key in `Dependencies` only coalescing would work. @pbrisbin wasn't that your use case?

eborden:
@cdparks couldn't coalescing take place in `genEntityKey` if we just passed the record to it as well? Why the other machinery?

```
genEntityKey record =
  CompositeKey
    (foo record) -- extracted
    (bar record) -- extracted
      <$> arbitrary -- some additional arbitrary value
```

eborden:
Oh I see why you are differentiating. I think I'd rather keep `embedkey` as an internal implementation for coherence (anyway) and just keep everything in `genEntityKey`.

cdparks:
You then need to change the type of the `genEntityKey` in a way that's easy to ignore:

```haskell
genEntityKey :: Thing -> Gen (Maybe (Entity Thing))
genEntityKey thing@Thing{..} =
  pure $ Just $ Entity (ThingKey foo bar) thing -- <-- works, we just pulled stuff out
  
genEntityKey :: Blah -> Gen (Maybe (Entity Blah))
genEntityKey blah@Blah{..} = do
  yarp <- someArbitraryComputation
  pure $ Entity (BlahKey foo bar yarp) blah -- <-- wrong, failed to update yarp in blah
```

cdparks:
I guess here's the thing that bugs me about the current scenario. You have to do this:

```haskell
instance Arbitrary StarScoreSkillMapping where
  arbitrary = genericArbitrary

instance EntityKeyGen StarScoreSkillMapping where
  genEntityKey =
    Just <$> (StarScoreSkillMappingKey <$> arbitrary <*> arbitrary)

instance HasDependencies StarScoreSkillMapping where
  type Dependencies StarScoreSkillMapping = Only SkillId
```

But the work happening in `genEntityKey` has already been done in the `Arbitrary` instance for `StarScoreSkillMapping`. We can just yank those bits out!

```diff
  instance Arbitrary StarScoreSkillMapping where
    arbitrary = genericArbitrary

- instance EntityKeyGen StarScoreSkillMapping where
-   genEntityKey =
-     Just <$> (StarScoreSkillMappingKey <$> arbitrary <*> arbitrary)

  instance HasDependencies StarScoreSkillMapping where
    type Dependencies StarScoreSkillMapping = Only SkillId

+   extractKey StarScoreSkillMapping {..} =
+     Just $ StarScoreSkillMappingKey
+       starScoreSkillMappingStarScore
+       starScoreSkillMappingSkillId
```

cdparks:
In the scenario where the client does implement `genKey`, `embedKey` is still internal. We call it, not the client.

cdparks:
One of these days, I'm gonna bite the bullet and do this

```haskell
share [mkGraphula, mkPersist sqlSettings] [persistLowerCase|
MathNumberBasicsQuestion json sql=math_number_basics_questions
    Id                  UUID sqltype=uuid
    skillCode           NumberBasicsSkillCode
    answerType          NumberBasicsAnswerType
    questionDiagramName Text Maybe
    updatedAt           UTCTime
    position            Natural
    deriving Generic
|]
```

eborden:
But what about the case when you need to extract and you have randomness. Just adding `record` to the arguments of `genEntityKey` enables all this flexibility with little cost to the API.

```hs
  extractKey :: a -> Maybe (Key a) -- Pure, never need to call embedKey

  genKey :: Gen (Maybe (Key a)) -- Has access to Arbitrary, must call embedKey
```
vs
```hs
  genKey :: a -> Gen (Maybe (Key a)) -- Has access to Arbitrary and subsumes extractKey
```

cdparks:
I guess that's fine. I don't love having to pay for `embedKey` in the case where we don't need it, but I'd already convinced myself it wasn't that expensive

eborden:
Yeah `embedKey` should be exceedingly cheap compared to any of the `IO` that happens here. And it only needs to happen in cases where we have a custom `Key`.

eborden:
@pbrisbin that would be helpful to validating the API.

cdparks:
I would be surprised if the case where you need to extract _and_ have randomness exists. That would mean you have a node that has an embedded key and the `Arbitrary` instance for the node is insufficient.

cdparks:
Yes please @pbrisbin

pbrisbin:
Entity:

```hs
{-# LANGUAGE NoDeriveAnyClass #-}
{-# LANGUAGE QuasiQuotes #-}
{-# LANGUAGE TemplateHaskell #-}

module FrontRow.Entities.RlSkill
  ( module FrontRow.Entities.RlSkill
  )
where

import Prelude

import Data.Aeson
import Database.Persist
import Database.Persist.TH
import FrontRow.Core.SchoolGrade
import FrontRow.Entities.RlAttributes
import FrontRow.Entities.RlStandard
import FrontRow.Entities.RlStandardSet
import FrontRow.Json
import GHC.Generics (Generic)
import Graphula (EntityKeyGen(..), HasDependencies(..))
import Test.QuickCheck.Arbitrary.Generic

mkPersist sqlSettings [persistLowerCase|
RlSkill sql=rl_skills
  uspId UspId
  name Name
  shortName Name
  progressionOrder ProgressionOrder
  grade SchoolGrade
  standardSetId RlStandardSetId
  standardId RlStandardId
  Primary uspId standardSetId
  deriving Show Eq Generic
|]

instance ToJSON RlSkill where
  toJSON = genericToJSON $ aesonOptions $ Just "rlSkill"
  toEncoding = genericToEncoding $ aesonOptions $ Just "rlSkill"

instance Arbitrary RlSkill where
  arbitrary = genericArbitrary

instance EntityKeyGen RlSkill where
  genEntityKey = Just <$> arbitrary

instance HasDependencies RlSkill where
  type Dependencies RlSkill = (RlStandardSetId, RlStandardId)

instance {-# OVERLAPPING #-} Arbitrary (Key RlSkill) where
  arbitrary = RlSkillKey <$> arbitrary <*> arbitrary
```

```hs
-- Doesn't work
        skill <- nodeWith @RlSkill $ keys (standardSet, standard)

-- Also doesn't work
        skill <- nodeEditWith @RlSkill (keys (standardSet, standard))
          $ (persistFieldLens RlSkillStandardSetId .~ entityKey standardSet)
          . (persistFieldLens RlSkillStandardId .~ entityKey standard)

-- Workaround
  skill <- liftIO $ generate arbitrary
  insertEntity $ skill
    { rlSkillStandardSetId = entityKey standardSet
    , rlSkillStandardId = entityKey standard
    }
```

cdparks:
Ok, this is actually a really interesting case because the primary key is also the dependencies. We actually _can't_ generate a new random key and embed because then we'll have overwritten the dependencies!

If we have separate `extractKey` (pure) and `genKey` (impure) methods, we have this:

```haskell
instance HasDependencies RlSkill where
  type Dependencies RlSkill = (RlStandardSetId, RlStandardId)

  extractKey RLSkill{..} =
    Just $ RLSkillKey
      RLSkillUspId
      RlSkillStandardSetId
```

In the case where we only have `genKey`, we have:

```haskell
instance HasDependencies RlSkill where
  type Dependencies RlSkill = (RlStandardSetId, RlStandardId)

  genKey RLSkill{..} =
    pure $ Just $ RLSkillKey
      RLSkillUspId
      RlSkillStandardSetId
```

Note that the user could implement:

```haskell
instance HasDependencies RlSkill where
  type Dependencies RlSkill = (RlStandardSetId, RlStandardId)

  genKey _ = Just <$> arbitrary
```

But now the dependencies have been erased!

So, yeah, I maybe need to think about if there's a way to make sure dependency fields are obviously untouchable.

eborden:
@cdparks you seem to be hunting for correct by construction. However exposing `genKey` and `extractKey` as separate APIs still doesn't give you that. An individual could easily write `genKey` in a place where they should have written `extractKey`. By packing everything in `genKey` you get a solution that generically subsumes `extractKey` and is a single point of API surface. Additionally what happens if I implement both `genKey` and `extractKey`?

cdparks:
`extractKey` takes precendence. We could warn on both implemented or even throw. This is the order of updates that I think is at least correct:

```haskell
nodeEditWith
  :: forall a m
   . GraphulaContext m '[a]
  => Dependencies a
  -> (a -> a)
  -> m (Entity a)
nodeEditWith dependencies edits = 10 `attemptsToInsertWith` do
  initialNode <- generateNode
  let
    edited = edits initialNode
    finalNode = edited `dependsOn` dependencies -- <- overwrites edits (as before)
    mExtractedKey = extractKey finalNode
  mKey <-
    case mExtractedKey of
      Nothing -> genExternalKey
      Just extractedKey ->
        mExternalKey <- genExternalKey
        maybe
          (pure $ Just extractedKey)
          (throwIO . usefulErrorMessage)
          mExternalKey
  logNode finalNode
  pure (mKey, finalNode)
```

This
1. Doesn't allow `genExternalKey` to override fields in the record (because that could break dependencies)
2. Requires that embedded keys use `Arbitrary` in the `Arbitrary` instance for the record itself
3. Logs the node that we actually intend to store
4. Explodes if you've implemented both. Again, I don't have a smart way to prevent this at compile time because `persistent` does not distinguish these cases at compile time.

cdparks:
> you seem to be hunting for correct by construction

Yeah, i don't think i can get this without template haskell

cdparks:
@pbrisbin since your dependencies fully determine your primary key, I don't think you even need to implement `genEntityKey`. Do your tests work if you change your instance to

```haskell
instance EntityKeyGen RlSkill
```

And use

```haskell
skill <- nodeWith @RlSkill $ keys (standardSet, standard)
```

?

pbrisbin:
Will check. Going to be a bit of recompile to get an answer; stand by.

eborden:
Oh man, not generating a key should work in this circumstance :face-palm:

cdparks:
Bwaaaawmp

pbrisbin:
Yup, looks like it worked. :man_facepalming:

cdparks:
Ok, so, should we just discard this for now? I'm not sure if we came up with something better than the (admittedly confusing) status quo

eborden:
Yeah, this will at least leave a document of some ideas if this comes up again.

PR #29, opened 2020-02-06
Simplify node API while expanding the set of supported relationships

cdparks:
**Breaking change**

This is a draft solution to an issue @pbrisbin noted in slack:

> so, math_question_content_{lang} 's own primary key id field is also the FK to math_questions(id)... Graphula cannot deal with this

Previously, we had (constraints elided)

```haskell
node         :: (Dependencies a ~ ())             => m (Entity a)
nodeEdit     :: (Dependencies a ~ ()) => (a -> a) -> m (Entity a)
nodeWith     ::  Dependencies a                   -> m (Entity a)
nodeEditWith ::  Dependencies a       -> (a -> a) -> m (Entity a)
```

<details>
  <summary>
    Click here for oldest stuff
  </summary>

We wanted to add a way to override an entity's primary key at the call site, but that would necessarily expand the set of bindings we'd have to export. So, I've changed the API to just:

```haskell
root :: (Dependencies a ~ ()) => NodeOptions a -> m (Entity a)
node ::  Dependencies a       -> NodeOptions a -> m (Entity a)
```

`NodeOptions a` is an opaque data type constructed by:

```haskell
mempty   ::                NodeOptions a
edit     :: (a -> a)    -> NodeOptions a
suchThat :: (a -> Bool) -> NodeOptions a
keyed    :: Key a       -> NodeOptions a
```

Note the `suchThat` constructor borrowed from @eborden 's PR #23. Multiple `NodeOptions a` values can be combined with the `Semigroup` operator `(<>)`:

```haskell
Entity schoolId _ <- root @School
  $ keyed (SchoolKey 1)
  <> edit (\s -> s { schoolName = "The School" })
```

One can avoid the extra parenthesization around the lambda with `do`:

```haskell
Entity schoolId _ <- root @School
  $ keyed (SchoolKey 1)
  <> do edit $ \s -> s { schoolName = "The School" }
```

Or by using `mconcat`:

```haskell
Entity schoolId _ <- root @School $ mconcat
  [ keyed (SchoolKey 1)
  , edit $ \s -> s { schoolName = "The School" }
  ]
```

I don't expect the `keyed` option to be used very often. The common case will be only passing `mempty` or `edit`.

For Pat's specific issue, we would have:

```haskell
instance HasDependencies MathQuestionContentEn
instance EntityKeyGen MathQuestionContentEn
...
Entity questionId _ <- node (only nf3SkillId) mempty
Entity contentId _ <- root $ keyed $ coerce questionId -- newtype
...
```
</details>

<details>
  <summary>
    Click here for older stuff
  </summary>

Specify `KeySource` in `HasDependencies`

This simplifies the `node` API even further. For a regular entity, a client now declares:

```haskell
mkPersist sqlSettings [persistLowerCase|
  Teacher
    name Text
    deriving Eq Show Generic

  Student
    name Text
    teacherId TeacherId Maybe
    deriving Eq Show Generic
|]

instance HasDependencies Teacher where
  type Dependencies Teacher = ()                 -- default
  type KeySource Teacher = 'GenerateKey 'Default -- default

instance HasDependencies Student where
  type Dependencies Student = Only (Maybe TeacherId)
  type KeySource Student = 'GenerateKey 'Default -- default
```

Using this information, we can statically assert that these values will always get their key from the database. We can use `node` as before to create entities:

```haskell
Entity teacherId _ <- node @Teacher $ edit $ \t -> t { teacherName = "Joe" }
Entity studentId _ <- node @Student (only $ Just teacherId) mempty
```

If we have a type with key that we'd rather generate with `Arbitrary` (e.g. a `UUID`), we can do like so:

```haskell
mkPersist sqlSettings [persistLowerCase|
  MathQuestion
    Id QuestionIdText
    skillId SkillId
    answerType AnswerType
    deriving Eq Show Generic
|]

deriving instance {-# OVERLAPPING #-} Arbitrary (Key MathQuestion)

instance HasDependencies MathQuestion where
  type Dependencies MathQuestion = Only SkillId
  type KeySource MathQuestion = 'GenerateKey 'Arbitrary
```

Again, `node` works as before, only we'll use `Arbitrary` ahead of time to create a key:

```haskell
Entity questionId _ <- node @MathQuestion (only skillId) mempty
```

Finally, if we have a type whose key should always be externally supplied, we do like so:

```haskell
mkPersist sqlSettings [persistLowerCase|
  MathQuestionContent
    Id QuestionIdText
    content Content
    deriving Eq Show Generic
|]

instance HasDependencies MathQuestionContent where
  type Dependencies MathQuestionContent = () -- default
  type KeySource MathQuestionContent = 'SpecifyKey
```

`node` will now require an extra argument following the type's dependencies:

```haskell
Entity questionId _ <- node @MathQuestion (only skillId) mempty
Entity contentId _ <- node @MathQuestionContent () (MathQuestionContentKey questionId) mempty
```

This uses some magic, but it's hopefully mostly hidden.

Also `GenEntityKey` is dead. `'GenerateKey 'Arbitrary` makes it redundant.

</details>

**NEWEST STUFF**

Specify `KeySource` in `HasDependencies` and split `node` into `node` and `nodeKeyed` (constraints elided):

```haskell
node      ::          Dependencies a -> NodeOptions a -> (m (Entity a))
nodeKeyed :: Key a -> Dependencies a -> NodeOptions a -> (m (Entity a))
```

For a regular entity, a client now declares:

```haskell
mkPersist sqlSettings [persistLowerCase|
  Teacher
    name Text
    deriving Eq Show Generic

  Student
    name Text
    teacherId TeacherId Maybe
    deriving Eq Show Generic
|]

instance HasDependencies Teacher where
  type Dependencies Teacher = ()           -- default
  type KeySource Teacher = 'SourceDefault  -- default

instance HasDependencies Student where
  type Dependencies Student = Only (Maybe TeacherId)
  type KeySource Student = 'SourceDefault  -- default
```

Using this information, we can statically assert that these values will always get their key from the database. We can use `node` as before to create entities:

```haskell
Entity teacherId _ <- node @Teacher $ edit $ \t -> t { teacherName = "Joe" }
Entity studentId _ <- node @Student (only $ Just teacherId) mempty
```

If we have a type with key that we'd rather generate with `Arbitrary` (e.g. a `UUID`), we can do like so:

```haskell
mkPersist sqlSettings [persistLowerCase|
  MathQuestion
    Id QuestionIdText
    skillId SkillId
    answerType AnswerType
    deriving Eq Show Generic
|]

deriving instance {-# OVERLAPPING #-} Arbitrary (Key MathQuestion)

instance HasDependencies MathQuestion where
  type Dependencies MathQuestion = Only SkillId
  type KeySource MathQuestion = 'SourceArbitrary
```

Again, `node` works as before, only we'll use `Arbitrary` ahead of time to create a key:

```haskell
Entity questionId _ <- node @MathQuestion (only skillId) mempty
```

Finally, if we have a type whose key should always be externally supplied, we do like so:

```haskell
mkPersist sqlSettings [persistLowerCase|
  MathQuestionContent
    Id QuestionIdText
    content Content
    deriving Eq Show Generic
|]

instance HasDependencies MathQuestionContent where
  type KeySource MathQuestionContent = 'SourceExternal
```

Now, the user cannot use `node` to create a value of type `MathQuestionContent`. This code:

```haskell
Entity questionId _ <- node @MathQuestion (only skillId) mempty
Entity contentId _ <- node @MathQuestionContent () mempty
```

will fail with an error message:

```plaintext
• Cannot generate a value of type ‘MathQuestionContent’ using ‘node’ since

    instance HasDependencies MathQuestionContent where
      type KeySource MathQuestionContent = 'SourceExternal

  Possible fixes include:
  • Use ‘nodeKeyed’ instead of ‘node’
  • Change ‘KeySource MathQuestionContent’ to 'SourceDefault or 'SourceArbitrary
    |
205 | Entity contentId _ <- node @MathQuestionContent () mempty
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^
```

As suggested by the error message, the user must instead use `nodeKeyed`:

```haskell
Entity questionId _ <- node @MathQuestion (only skillId) mempty
Entity contentId _ <- nodeKeyed @MathQuestionContent (MathQuestionContentKey questionId) () mempty
```

`GenEntityKey` has been repurposed internally, and should no longer be implemented by the user.

### Alternatives

I briefly explored putting the dependencies inside `NodeOptions a`, but then you need to make sure there's no way _not_ to pass dependencies. [This requires more magic than I'm comforable with](https://gist.github.com/cdparks/08938203705df523b4a7102880030639).

A simpler model uses function application as the `(<>)` operator and makes the dependencies constructor (`with`) the only option that takes no arguments, e.g.:

```haskell
with     :: Dependencies a -> NodeOptions a
edit     :: (a -> a)       -> NodeOptions a -> NodeOptions a
suchThat :: (a -> Bool)    -> NodeOptions a -> NodeOptions a
keyed    :: Key a          -> NodeOptions a -> NodeOptions a

-- Then:

Entity schoolId _ <- root @School
Entity teacherId _ <- node @Teacher
  $ with (only $ Just schoolId)
  & keyed (TeacherKey 1)
  & edit (\s -> s { schoolName = "The School" })
```

Now we don't have to pass `mempty` everywhere, but we still need to over-parenthesize `edit` lambdas.

### DIFF
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -47,7 +47,7 @@ module Graphula
   , NodeOptions
   , edit
   , suchThat
-  , primaryKey
+  , keyed
     -- ** Declaring Dependencies
   , HasDependencies(..)
     -- *** Singular Dependencies
@@ -463,16 +463,16 @@ type GraphulaNode m a =
 --
 -- @
 -- a1 <- root @A mempty
--- a2 <- root @A $ primaryKey $ AKey 1
+-- a2 <- root @A $ keyed $ AKey 1
 -- a3 <- root @A $ edit $ \a -> a { someField = True }
 -- a4 <- root @A $ suchThat $ (== True) . someField
--- a5 <- root @A $ primaryKey (AKey 2) <> edit (\a -> a { someField = True })
--- a6 <- root @A $ primaryKey (AKey 3) <> do edit $ \a -> a { someField = True }
+-- a5 <- root @A $ keyed (AKey 2) <> edit (\a -> a { someField = True })
+-- a6 <- root @A $ keyed (AKey 3) <> do edit $ \a -> a { someField = True }
 -- @
 --
 data NodeOptions a = NodeOptions
   { nodeOptionsEdit :: Kendo Maybe a
-  , nodeOptionsPrimaryKey :: Last (Gen (Maybe (Key a)))
+  , nodeOptionsKeyed :: Last (Gen (Maybe (Key a)))
   }
   deriving (Generic)
 
@@ -521,12 +521,19 @@ suchThat f = mempty
 -- | Override any automatic key generation with the specified key
 --
 -- @
--- a <- root @A $ primaryKey $ AKey 1
+-- a <- root @A $ keyed $ AKey 1
 -- @
 --
-primaryKey :: Key a -> NodeOptions a
-primaryKey key = mempty
-  { nodeOptionsPrimaryKey = Last $ Just $ pure $ Just key
+-- If this is called more than once, the last key is used. E.g.,
+-- @AKey 2@ would be used below:
+--
+-- @
+-- a <- root @A $ keyed (AKey 1) <> keyed (AKey 2)
+-- @
+--
+keyed :: Key a -> NodeOptions a
+keyed key = mempty
+  { nodeOptionsKeyed = Last $ Just $ pure $ Just key
   }
 
 -- | Generate a value that does not have any dependencies
@@ -545,7 +552,7 @@ root = node ()
   dependency data is inserted after any editing operations.
 
   > node @Dog (ownerId, veterinarianId) mempty
-  > node @Dog (ownerId, vererinarianId) $ primaryKey $ DogKey 1
+  > node @Dog (ownerId, vererinarianId) $ keyed $ DogKey 1
   > node @Dog (ownerId, vererinarianId) $ edit $ \dog ->
   >   dog {name = "fido"}
 -}
@@ -560,7 +567,7 @@ node dependencies NodeOptions{..} = attempt 100 10 $ do
   for (appKendo nodeOptionsEdit initial) $ \edited -> do
     let hydrated = edited `dependsOn` dependencies
     logNode hydrated
-    mKey <- liftIO $ generate $ fromMaybe genEntityKey $ getLast nodeOptionsPrimaryKey
+    mKey <- liftIO $ generate $ fromMaybe genEntityKey $ getLast nodeOptionsKeyed
     pure (mKey, hydrated)
 
 attempt
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -30,6 +30,7 @@
 {-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
@@ -40,10 +41,12 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeFamilyDependencies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 module Graphula
   ( -- * Graph Declaration
     node
+  , nodeKeyed
   , GraphulaNode
   , GraphulaContext
     -- ** Node options
@@ -53,7 +56,6 @@ module Graphula
     -- * Declaring Dependencies and key source
   , HasDependencies(..)
   , KeySourceType(..)
-  , KeyStrategy(..)
     -- ** Singular Dependencies
   , Only(..)
   , only
@@ -124,13 +126,14 @@ import Database.Persist.Sql (SqlBackend)
 import Generics.Eot (Eot, HasEot, fromEot, toEot)
 import GHC.Exts (Constraint)
 import GHC.Generics (Generic)
+import GHC.TypeLits (TypeError, ErrorMessage(..))
 import Graphula.Internal
 import System.Directory (createDirectoryIfMissing, getTemporaryDirectory)
 import System.IO (Handle, IOMode(..), hClose, openFile)
 import System.IO.Temp (openTempFile)
 import Test.HUnit.Lang
   (FailureReason(..), HUnitFailure(..), formatFailureReason)
-import Test.QuickCheck (Arbitrary(..), Gen, generate)
+import Test.QuickCheck (Gen, Arbitrary(..), generate)
 import UnliftIO.Exception
   (Exception, SomeException, bracket, catch, mask, throwIO)
 
@@ -412,6 +415,8 @@ class NoConstraint a where
 
 instance NoConstraint a where
 
+
+
 class HasDependencies a where
   -- | A data type that contains values to be injected into @a@ via
   -- `dependsOn`. The default generic implementation of `dependsOn` supports
@@ -428,16 +433,17 @@ class HasDependencies a where
   -- This can be
   --
   -- @
-  -- 'GenerateKey 'Default   -- automatically generate keys from the database
-  -- 'GenerateKey 'Arbitrary -- automatically generate keys using @'Arbitrary'@
-  -- 'SpecifyKey             -- explicitly pass keys to @'node'@
+  -- 'SourceDefault   -- automatically generate keys from the database
+  -- 'SourceArbitrary -- automatically generate keys using @'Arbitrary'@
+  -- 'SourceExternal  -- explicitly pass a key using @'nodeKeyed'@
   -- @
   --
-  -- Most types will use @'GenerateKey Default'@ or @'GenerateKey Arbitrary'@.
-  -- Only use @'SpecifyKey'@ if the key for a value is always defined
+  -- Most types will use @'SourceDefault'@ or @'SourceArbitrary'@. Only
+  -- use @'SourceExternal'@ if the key for a value is always defined
   -- externally.
+  --
   type KeySource a :: KeySourceType
-  type instance KeySource a = 'GenerateKey 'Default
+  type instance KeySource a = 'SourceDefault
 
   -- | Assign values from the 'Dependencies' collection to a value.
   -- 'dependsOn' must be an idempotent operation.
@@ -461,19 +467,80 @@ class HasDependencies a where
         (toEot a)
         (toEot dependencies)
 
--- | Source of keys
 data KeySourceType
-  = GenerateKey KeyStrategy
-  -- ^ Generate keys using the specified @'Strategy'@
-  | SpecifyKey
+  = SourceDefault
+  -- ^ Generate keys using the database's @DEFAULT@ strategy
+  | SourceArbitrary
+  -- ^ Generate keys using the @'Arbitrary'@ instance for the @'Key'@
+  | SourceExternal
   -- ^ Always explicitly pass an external key
 
--- | How to implicitly generate keys
-data KeyStrategy
-  = Default
-  -- ^ Let the database generate keys
-  | Arbitrary
-  -- ^ Use '@Arbitrary'@ to generate keys
+-- | Handle key generation for @'SourceDefault'@ and @'SourceArbitrary'@
+--
+-- Ths could be a single-parameter class, but carrying the @a@ around
+-- lets us give a better error message when @'node'@ is called instead
+-- of @'nodeKeyed'@.
+--
+class GenEntityKey (s :: KeySourceType) a where
+  type KeyConstraint s a :: Constraint
+  genEntityKey :: KeyConstraint s a => Gen (Maybe (Key a))
+
+instance GenEntityKey 'SourceDefault a where
+  type KeyConstraint 'SourceDefault a = NoConstraint a
+  genEntityKey = pure Nothing
+
+instance GenEntityKey 'SourceArbitrary a where
+  type KeyConstraint 'SourceArbitrary a = Arbitrary (Key a)
+  genEntityKey = Just <$> arbitrary
+
+type family Quote t where
+  Quote t = 'Text "‘" ':<>: t ':<>: 'Text "’"
+
+-- | Explicit instance for @'SourceExternal'@ to give an actionable error message
+--
+-- Rendered:
+--
+-- @
+-- Cannot generate a value of type ‘X’ using ‘node’ since
+--
+--   instance HasDependencies X where
+--     type KeySource X = 'SourceExternal
+--
+-- Possible fixes include:
+-- • Use ‘nodeKeyed’ instead of ‘node’
+-- • Change ‘KeySource X’ to 'SourceDefault or 'SourceArbitrary
+-- @
+--
+instance TypeError
+  ( 'Text "Cannot generate a value of type "
+    ':<>: Quote ('ShowType a)
+    ':<>: 'Text " using "
+    ':<>: Quote ('Text "node")
+    ':<>: 'Text " since"
+    ':$$: 'Text ""
+    ':$$: 'Text "  instance HasDependencies "
+    ':<>: 'ShowType a
+    ':<>: 'Text " where"
+    ':$$: 'Text "    "
+    ':<>: 'Text "type KeySource "
+    ':<>: 'ShowType a
+    ':<>: 'Text  " = "
+    ':<>: 'ShowType 'SourceExternal
+    ':$$: 'Text ""
+    ':$$: 'Text "Possible fixes include:"
+    ':$$: 'Text "• Use "
+    ':<>: Quote ('Text "nodeKeyed")
+    ':<>: 'Text " instead of "
+    ':<>: Quote ('Text "node")
+    ':$$: 'Text "• Change "
+    ':<>: Quote ('Text "KeySource " ':<>: 'ShowType a)
+    ':<>: 'Text " to "
+    ':<>: 'Text "'SourceDefault"
+    ':<>: 'Text " or "
+    ':<>: 'Text "'SourceArbitrary"
+  ) => GenEntityKey 'SourceExternal a where
+  type KeyConstraint 'SourceExternal a = NoConstraint a
+  genEntityKey = error "unreachable"
 
 data GenerationFailure
   = GenerationFailureMaxAttemptsToConstrain TypeRep
@@ -493,7 +560,6 @@ type GraphulaNode m a
     , Typeable a
     )
 
-
 {-|
   Generate a value with data dependencies. This leverages
   'HasDependencies' to insert the specified data in the generated value. All
@@ -503,25 +569,57 @@ type GraphulaNode m a
   > node @Dog (ownerId, veterinarianId) $ edit $ \dog ->
   >   dog {name = "fido"}
 
-  A value that has an externally managed key will always require an extra argument
-
-  > someKey <- generateKey
-  > node @Cat (ownerId, veterinarianId) someKey mempty
-  > anotherKey <- generateKey
-  > node @Cat (ownerId, veterinarianId) anotherKey $ edit $ \cat ->
-  >   cat {name = "milo"}
+  A value that has an externally managed key must use @'nodeKeyed'@ instead.
 -}
 node
-  :: forall a s m r c
-   . ( GraphulaContext m '[a]
-     , KeySource a ~ s
-     , ResolveArgs s
-     , Args s m a ~ (Dict s c -> r)
-     , c
-     )
+  :: forall a s m
+  . ( GraphulaContext m '[a]
+    , KeySource a ~ s
+    , GenEntityKey s a
+    , KeyConstraint s a
+    )
   => Dependencies a
-  -> r
-node deps = resolveArgs @s @a @m deps Dict
+  -> NodeOptions a
+  -> (m (Entity a))
+node = nodeImpl $ genEntityKey @s @a
+
+{-|
+  Generate a value with data dependencies given an externally managed
+  key. This leverages 'HasDependencies' to insert the specified data
+  in the generated value. All dependency data is inserted after any
+  editing operations.
+
+  > someKey <- generateKey
+  > node @Cat someKey (ownerId, veterinarianId) mempty
+  > anotherKey <- generateKey
+  > node @Cat anotherKey (ownerId, veterinarianId) $ edit $ \cat ->
+  >   cat {name = "milo"}
+
+  A value that has an automatically managed key may use @'node'@ instead.
+-}
+nodeKeyed
+  :: forall a m
+  . GraphulaContext m '[a]
+  => Key a
+  -> Dependencies a
+  -> NodeOptions a
+  -> (m (Entity a))
+nodeKeyed key = nodeImpl $ pure $ Just key
+
+nodeImpl
+  :: forall a m
+  . GraphulaContext m '[a]
+  => Gen (Maybe (Key a))
+  -> Dependencies a
+  -> NodeOptions a
+  -> (m (Entity a))
+nodeImpl genKey dependencies NodeOptions {..} = attempt 100 10 $ do
+  initial <- generateNode
+  for (appKendo nodeOptionsEdit initial) $ \edited -> do
+    let hydrated = edited `dependsOn` dependencies
+    logNode hydrated
+    mKey <- liftIO $ generate $ genKey
+    pure (mKey, hydrated)
 
 -- | Modify the node after it's been generated
 --
@@ -530,7 +628,9 @@ node deps = resolveArgs @s @a @m deps Dict
 -- @
 --
 edit :: (a -> a) -> NodeOptions a
-edit f = mempty { nodeOptionsEdit = Kendo $ Just . f }
+edit f = mempty
+  { nodeOptionsEdit = Kendo $ Just . f
+  }
 
 -- | Require a node to satisfy the specified predicate
 --
@@ -539,7 +639,9 @@ edit f = mempty { nodeOptionsEdit = Kendo $ Just . f }
 -- @
 --
 suchThat :: (a -> Bool) -> NodeOptions a
-suchThat f = mempty { nodeOptionsEdit = Kendo $ \a -> a <$ guard (f a) }
+suchThat f = mempty
+  { nodeOptionsEdit = Kendo $ \a -> a <$ guard (f a)
+  }
 
 -- | Options for generating an individual node
 --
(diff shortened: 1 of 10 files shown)


graphula-core/src/Graphula.hs
```diff
@@ -31,17 +30,21 @@
...
 {-# LANGUAGE TypeOperators #-}
 
 module Graphula
   ( -- * Graph Declaration
-    node
-  , nodeEdit
-  , nodeWith
-  , nodeEditWith
+    root
+  , node
```
eborden: Probably want `GraphulaNode` and `GraphulaContext` up here.

graphula-core/src/Graphula.hs
```diff
@@ -31,17 +30,21 @@
...
+    -- * Node options
```
eborden: Should this be an h2 `**`

graphula-core/src/Graphula.hs
```diff
@@ -445,69 +454,124 @@ type GraphulaNode m a =
...
+-- a1 <- root @A mempty
+-- a2 <- root @A $ primaryKey $ AKey 1
+-- a3 <- root @A $ edit $ \a -> a { someField = True }
+-- a4 <- root @A $ suchThat $ (== True) . someField
+-- a5 <- root @A $ primaryKey (AKey 2) <> edit (\a -> a { someField = True })
+-- a6 <- root @A $ primaryKey (AKey 3) <> do edit $ \a -> a { someField = True }
+-- @
+--
+data NodeOptions a = NodeOptions
+  { nodeOptionsEdit :: Endo (Maybe a)
```
eborden: To avoid having to pass a redundant `Just a` to the starting function this could be kleisli composition for `<>`.

pbrisbin:
Going to review the code now, some thoughts just based on the PR Description:

First, :heart: looks dope.

> but then you need to make sure there's no way not to pass dependencies

Possibly not surprisingly, I don't feel the need to "make sure" of this with type magic. If there existed an API where dependencies were optional, assuming omitting them meant `type Dependences = ()`, that sounds pretty nice. Now we only need `node`, not `root`.

> A simpler model uses function application

Coming at this strictly from _usage_, I have no problem parenthesizing the lamba (most often, I'll be using lens here, which means there is no lambda anyway), and actually prefer that over using `do` (confusing) or `mconcat` (noisy) to avoid it. Given that, the `<>` vs `&` usage is only a minor aesthetic difference that I could take or leave.

I wonder if you meant "simpler" in terms of this usage UX or the internals/implementation/types. If one's internals are simpler than the other, I vote for the simpler one. If it's just your opinion of the usage UX, I defer to you which way to go.

graphula-core/src/Graphula.hs
```diff
@@ -31,17 +30,21 @@
...
-  , nodeEdit
-  , nodeWith
-  , nodeEditWith
+    root
+  , node
+    -- * Node options
+  , NodeOptions
+  , edit
+  , suchThat
+  , primaryKey
```
pbrisbin: What about `withKey` or `keyed`?

`suchThat` has a nice sentence-ness to it, and `edit` is a sort of "do this" tense. So either of `withKey` or `keyed` feels in the same spirit to me. `withKey` might be a nice symmetry for `with`. IMO, `primaryKey` is just `key` but with false descriptiveness added, perhaps to work around a conflict. Or are there non-primary keys in this context worth distinguishing from? I'd be fine with `key` too, but I do like `withKey` or `keyed` better.

graphula-core/src/Graphula.hs
```diff
@@ -445,69 +454,124 @@ type GraphulaNode m a =
...
-node
-  :: forall a m . (GraphulaContext m '[a], Dependencies a ~ ()) => m (Entity a)
-node = nodeWith ()
-
-attemptsToInsertWith
+node dependencies NodeOptions{..} = attempt 100 10 $ do
+  initial <- generateNode
+  for (appEndo nodeOptionsEdit $ Just initial) $ \edited -> do
+    let hydrated = edited `dependsOn` dependencies
+    logNode hydrated
```
eborden: Is there an issue with `logNode` being here? If a node fails to insert, will `logNode` put multiple items in the log?

eborden:
The big question, how do we transition?

pbrisbin:
```hs
module Graphula.Compat where

nodeWith x = node $ with x

nodeEditWith x d = node $ with d <> edit f
```

?

eborden:
`node` breaks though.

pbrisbin:
Still a major version bump, and `node` itself is still [breaking-]changed (right)? But at least having this import handy could make downstream apps lives easier.

eborden:
It could certainly expedite a lot.

pbrisbin:
We *could* do,

```hs
module Graphula.Compat
  ( node
  , nodeWith
  , nodeEditWith
  , module Graphula
  )
where

import Graphula hiding (node)
import qualified Graphula as G

node = {- something -}

nodeWith x = G.node $ with x

nodeEditWith x f = G.node $ with x <> edit f
```

But personally, I think a breaking `node` wouldn't be too much fallout in our own apps, we're *much* heavier users of the `*With` varieties, for which that first `.Compat` would be invaluable, IMO.

eborden:
Agreed, `node` is used much less. When discussing designs @cdparks and I originally talked about how `node () mempty` wouldn't be a huge burden because dependency-less generation is not the norm.

graphula-core/src/Graphula.hs
cdparks: This is how it was working before. I wasn't sure if I should change it. Also, if you do fail to insert, you probably want to have logged something.

graphula-core/src/Graphula.hs
cdparks: I kind of like `keyed` best

cdparks:
I don't love compat modules, I'd rather just do the big-bang fix myself. I can probably automate most of it.

pbrisbin:
> I don't love compat modules

Can you say more? What's the downside (both generally, and in this case where it's likely to be a relatively small and write-once module, IMO)?

> I'd rather just do the big-bang fix myself. I can probably automate most of it.

Are you speaking as us at freckle knowing how `sed`-able our current usages are, or are you speaking as any "reasonable user" of this library?

If you're willing to convert our codebase right away, then for sure go for it -- with or without us adding some sort of `.Compat` module to make things easier other users.

Or do we just sort of assume we're the only users of this library?

graphula-core/src/Graphula.hs
eborden: > This is how it was working before. 

Yup, but it might have been broken then too.

> Also, if you do fail to insert, you probably want to have logged something.

The log is for replay and inspection of the graph when a test fails. So you don't want data in there that isn't used.

graphula-core/src/Graphula.hs
```diff
@@ -31,62 +30,82 @@
...
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 
 module Graphula
-  ( -- * Graph Declaration
-    node
-  , nodeEdit
-  , nodeWith
-  , nodeEditWith
+  ( -- ** Graph Declaration
```
eborden: Oh, I didn't mean all these should be h2, just that the "Node Options" might be best placed under "Graph Declaration".

graphula-core/src/Graphula.hs
```diff
@@ -429,85 +446,156 @@ class HasDependencies a where
...
-  :: forall a m . GraphulaContext m '[a] => Dependencies a -> m (Entity a)
-nodeWith = flip nodeEditWith id
+instance Monad m => Semigroup (Kendo m a) where
+  (<>) = coerce @((a -> m a) -> (a -> m a) -> a -> m a) (<=<)
+  {-# INLINE (<>) #-}
 
-{-|
-  Generate and edit a value that does not have any dependencies.
+instance Monad m => Monoid (Kendo m a) where
+  mempty = coerce @(a -> m a) pure
```
eborden: This feels like a "because I can" use of `coerce` where `Kendo pure` is succinct and readable.

cdparks:
> Are you speaking as us at freckle

Yeah, this isn't on hackage or stackage, I would be shocked if anyone else is using it.

graphula-core/src/Graphula.hs
cdparks: Every use of `coerce` is a "because I can" :grinning:

graphula-core/src/Graphula.hs
cdparks: Ah, I wasn't sure. I didn't know if Haddocks are supposed to obey the "only one H1 thing in the whole page" that applies to regular HTML

cdparks:
Did a big 'ol update. More magic, but "I need to use an external key" is now a compile error if you forget to do it instead of a runtime error. Also, it kills `GenEntityKey`

graphula-core/src/Graphula.hs
```diff
@@ -429,85 +461,230 @@ class HasDependencies a where
         (toEot a)
         (toEot dependencies)
 
-data GenerationFailure =
-  GenerationFailureMaxAttempts TypeRep
+-- | Source of keys
+data KeySourceType
+  = GenerateKey KeyStrategy
+  -- ^ Generate keys using the specified @'Strategy'@
+  | SpecifyKey
```
cdparks: I could do with better names here, I've also had

```haskell
data KeySourceType
  = Implicit KeyStrategy
  | Explicit
```

and

```haskell
data KeySourceType
  = Generate KeyStrategy
  | External
```

graphula-core/src/Graphula.hs
```diff
@@ -429,85 +461,230 @@ class HasDependencies a where
...
+-- @'NodeOptions'@ can be created and combined with the Monoidal
+-- operations @'(<>)'@ and @'mempty'@.
+--
+-- @
+-- a1 <- node @A () mempty
+-- a2 <- node @A () $ edit $ \a -> a { someField = True }
+-- a3 <- node @A () $ suchThat $ (== True) . someField
+-- @
+--
+newtype NodeOptions a = NodeOptions
```
cdparks: I still kept `NodeOptions` for `edit`/`suchThat`, and now we have an extension point for other optional bits.

graphula-core/test/README.lhs
```diff
@@ -15,23 +16,27 @@ Graphula is a simple interface for generating persistent data and linking its de
...
 import Database.Persist.Arbitrary ()
+import Database.Persist (Entity(..))
 import Database.Persist.Sqlite
 import Database.Persist.TH
+import Data.Typeable
+import GHC.Generics (Generic)
+import Graphula
+import Graphula.UUIDKey
+import Test.Hspec
+import Test.QuickCheck hiding (suchThat)
```
cdparks: @eborden can you think of a better name than `suchThat`? Trying to avoid a name clash. Other options:

* `where_` (clashes with `Esqueleto` tho)
* `satisfying`
* `given`
* `iff`

pbrisbin: I vote for `satisfying`. I might toss in `ensuring` too, or `passing`, or `filtered`. `scoped`?

cdparks: Hmm, I think I like `ensure` or `expect` since we're using non-participle `edit`

eborden: I like `ensure`

image

cdparks:
Actually pretty good as long as you provide a type application for the node you want. In the degenerate case, if I just pass `node @T` for a bunch of things:

```haskell
inferTest :: IO ()
inferTest = runGraphulaT runDB $ do
  node @A
  node @B
  node @C
  node @D
  node @E
  node @F
```

I get:

```plaintext
    /home/ubuntu/code/graphula/graphula-core/test/README.lhs:266:3: error:
    • Couldn't match type ‘() -> NodeOptions A -> m0 (Entity A)’
                     with ‘GraphulaT (NoLoggingT (ResourceT IO)) IO a0’
      Expected type: GraphulaT (NoLoggingT (ResourceT IO)) IO a0
        Actual type: Dependencies A -> NodeOptions A -> m0 (Entity A)
    • Probable cause: ‘node’ is applied to too few arguments
        ...
    |
266 |   node @A
    |   ^^^^^^^

/home/ubuntu/code/graphula/graphula-core/test/README.lhs:267:3: error:
    • Couldn't match type ‘Only AId -> NodeOptions B -> m1 (Entity B)’
                     with ‘GraphulaT (NoLoggingT (ResourceT IO)) IO a1’
      Expected type: GraphulaT (NoLoggingT (ResourceT IO)) IO a1
        Actual type: Dependencies B -> NodeOptions B -> m1 (Entity B)
    • Probable cause: ‘node’ is applied to too few arguments
        ...
    |
267 |   node @B
    |   ^^^^^^^

/home/ubuntu/code/graphula/graphula-core/test/README.lhs:268:3: error:
    • Couldn't match type ‘(AId, BId)
                           -> NodeOptions C -> m2 (Entity C)’
                     with ‘GraphulaT (NoLoggingT (ResourceT IO)) IO a2’
      Expected type: GraphulaT (NoLoggingT (ResourceT IO)) IO a2
        Actual type: Dependencies C -> NodeOptions C -> m2 (Entity C)
    • Probable cause: ‘node’ is applied to too few arguments
        ...
    |
268 |   node @C
    |   ^^^^^^^

/home/ubuntu/code/graphula/graphula-core/test/README.lhs:269:3: error:
    • Couldn't match type ‘() -> NodeOptions D -> m3 (Entity D)’
                     with ‘GraphulaT (NoLoggingT (ResourceT IO)) IO a3’
      Expected type: GraphulaT (NoLoggingT (ResourceT IO)) IO a3
        Actual type: Dependencies D -> NodeOptions D -> m3 (Entity D)
    • Probable cause: ‘node’ is applied to too few arguments
        ...
    |
269 |   node @D
    |   ^^^^^^^

/home/ubuntu/code/graphula/graphula-core/test/README.lhs:270:3: error:
    • Couldn't match type ‘()
                           -> Key E -> NodeOptions E -> m4 (Entity E)’
                     with ‘GraphulaT (NoLoggingT (ResourceT IO)) IO a4’
      Expected type: GraphulaT (NoLoggingT (ResourceT IO)) IO a4
        Actual type: Dependencies E
                     -> Key E -> NodeOptions E -> m4 (Entity E)
    • Probable cause: ‘node’ is applied to too few arguments
        ...
    |
270 |   node @E
    |   ^^^^^^^

/home/ubuntu/code/graphula/graphula-core/test/README.lhs:271:3: error:
    • Couldn't match type ‘() -> NodeOptions F -> m5 (Entity F)’
                     with ‘GraphulaT (NoLoggingT (ResourceT IO)) IO ()’
      Expected type: GraphulaT (NoLoggingT (ResourceT IO)) IO ()
        Actual type: Dependencies F -> NodeOptions F -> m5 (Entity F)
    • Probable cause: ‘node’ is applied to too few arguments
        ...
    |
271 |   node @F
    |   ^^^^^^^
```

Note that `ghc` expands `Dependencies _` and it correctly infers that you must pass a `Key E` for `@E`.

This is, I think, what I want. As long as you say `node @X`, `ghc` will tell you
1. What `Dependencies X` is
2. Whether it needs an explicit `Key X`
3. That you must pass `NodeOptions X`

It doesn't tell you that `NodeOptions X` can be `mempty`, but that would always be the case

eborden:
We have a number of cases where we don't pass type applications and return type polimorphism is used.

```hs

simpleSchoolWith
  :: ( GraphulaContext m '[District, School, Teacher, Student, Course]
     , GraphulaContext
         m
         '[CourseMembership, SchoolAdmin, SchoolAdminSchoolMembership]
     , MonadFail m
     )
  => (School -> School)
  -> (SchoolAdmin -> SchoolAdmin)
  -> (Teacher -> Teacher)
  -> m SimpleSchool
simpleSchoolWith schoolEdit schoolAdminEdit teacherEdit = do
  district1 <- node
  school1 <- nodeEditWith (onlyKey district1) schoolEdit
  schoolAdmin1 <- simpleSchoolAdminWith schoolAdminEdit
  teacher1 <- simpleTeacherWith (Just $ entityKey school1) teacherEdit
  students@[student1, student2, student3, student4] <- traverse
    (uncurry $ simpleStudent $ Just $ entityKey teacher1)
    [ ("Avery", "Jones")
    , ("Michael", "Benevento")
    , ("Jayden", "Carter")
    , ("Hunter", "Sandoval")
    ]
  course1 <-
    nodeEditWith (onlyKey teacher1) $ (persistFieldLens CourseArchived) .~ False
  course2 <-
    nodeEditWith (onlyKey teacher1) $ (persistFieldLens CourseArchived) .~ False
  for_ students $ nodeWith @CourseMembership . keys . (, course1)
  for_ students $ nodeWith @CourseMembership . keys . (, course2)
  void $ nodeWith @SchoolAdminSchoolMembership $ keys (schoolAdmin1, school1)
  pure SimpleSchool { .. }
```

As well I know there are cases where we infer from the the terminal node, like:

```hs
x <- node
y <- nodeWith $ onlyKey x
z <- nodeWith @Waffle $ keys (x, y)
```

Will those uses regress in this version?

cdparks:
Yes, `ghc` can't infer backwards from the terminal node in this API. The initial type application basically becomes required. Lemme see if I can recover better inference if I mush the key back into the `NodeOptions` but keep the knowledge of whether the key is required or not

cdparks:
Mm, that doesn't work even if you disallow `mempty` for keyed nodes because you can use `edit` to forget to pass a key.

This works and has good inference but you have to pay for an extra argument everywhere:

```haskell
node
  :: forall a m
   . GraphulaContext m '[a]
  => KeyParam (KeySource a) a
  -> Dependencies a
  -> NodeOptions a
  -> (m (Entity a))
...

simpleSpec :: IO ()
simpleSpec =
  runGraphulaT runDB $ do
    Entity aId _ <- node Serial () mempty
    Entity bId _ <- node Serial (only aId) mempty
    void $ node @C Serial (aId, bId) $ edit $ \n -> n { cC = "edited" }
    void $ node Random () mempty
    void $ node @E (Inject (EKey dId)) () mempty
```

cdparks:
Really gotta finish an RFC, so I'm gonna take a break from this and see if something else knocks loose in my brain later

eborden:
Is losing good type inference an acceptable cost.

> This function requires type applications.

eborden:
This is kind of the natural push and pull between Hindley Milner and dependent types. Dependent types ask you to provide more type information so they can synthesize more code for you.

cdparks:
I think we're maybe trying to do two thing at once:
1. Move `edit` & co into options
2. Avoid blow up of `nodeX` functions

However, moving `edit` into options frees us up to keep the `node` API the same size as before - that is, maybe we should have `nodeX` for different key strategies. I think I can probably mash the `Default` and `Arbitrary` strategy into `node` and then have `nodeKeyed` which always takes the extra key param

eborden:
I mean, I like the idea that configuration leads to how something is used. That avoids lots of bugs at the call site. I'm just trying to poke the design space and see where it leads us.

cdparks:
Ok, new idea - this splits `node` into `node` and `nodeKeyed`:

```haskell
node
  :: forall a s m
  . ( GraphulaContext m '[a]
    , KeySource a ~ s
    , GenEntityKey s a
    , KeyConstraint s a
    )
  => Dependencies a
  -> NodeOptions a
  -> (m (Entity a))

nodeKeyed
  forall a m
  . GraphulaContext m '[a]
  => Key a
  -> Dependencies a
  -> NodeOptions a
  -> (m (Entity a))
```

You can only call `node` for types that have:

```haskell
instance HasDependencies X where
  type KeySource = 'SourceDefault
```

Or

```haskell
instance HasDependencies X where
  type KeySource = 'SourceArbitrary
```

You must call `nodeKeyed` for types that have:

```haskell
instance HasDependencies X where
  type KeySource = 'SourceExternal
```

If you mess this up, you get a nice error message:

```plaintext
/home/ubuntu/code/graphula/graphula-core/test/README.lhs:205:21: error:
    • Cannot generate a value of type ‘E’ using ‘node’ since

        instance HasDependencies E where
          type KeySource E = 'SourceExternal

      Possible fixes include:
      • Use ‘nodeKeyed’ instead of ‘node’
      • Change ‘KeySource E’ to 'SourceDefault or 'SourceArbitrary
    • In a stmt of a 'do' block:
        Entity eId _ <- node @E (EKey dId) () mempty
      In the second argument of ‘($)’, namely
        ...

    |
205 |     Entity eId _ <- node @E (EKey dId) () mempty
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I don't prevent you from calling `nodeKeyed` with the other source type instances. I consider that a reasonable escape hatch, but if we wanted to, we could prevent it with:

```haskell
nodeKeyed
  forall a m
  . (GraphulaContext m '[a], KeySource s a ~ 'SourceExternal)
  => Key a
  -> Dependencies a
  -> NodeOptions a
  -> (m (Entity a))
```

I could be convinced to call `nodeKeyed` just `keyed`, but I thought we'd wanna retain the `node` bit.

Inference is good. Either a return type annotation here or a `@C` type application is enough here:

```haskell
_inferTest = runGraphulaT runDB $ do
  Entity aId _ <- node () mempty
  Entity bId _ <- node (only aId) mempty
  node @C (aId, bId) mempty
```

cdparks:
Description updated again

cdparks:
thoughts @eborden ?

eborden:
Sorry, I've been busy hunting other stuff down and partially trying to not distract you :rofl: 

I'll take a look at this tomorrow morning and hit you up with that sweet sweet feedback.

graphula-core/src/Graphula.hs
```diff
@@ -429,85 +467,234 @@ class HasDependencies a where
...
+  A value that has an externally managed key must use @'nodeKeyed'@ instead.
 -}
-nodeEditWith
-  :: forall a m
-   . GraphulaContext m '[a]
+node
+  :: forall a s m
+   . ( GraphulaContext m '[a]
+     , KeySource a ~ s
+     , GenEntityKey s a
```
eborden: So if I call this function with a `SouceExternal` type, then I will get a type error from the instance resolution?

graphula-core/src/Graphula.hs
```diff
@@ -429,85 +467,234 @@ class HasDependencies a where
...
+  -- ^ Always explicitly pass an external key
+
+-- | Handle key generation for @'SourceDefault'@ and @'SourceArbitrary'@
+--
+-- Ths could be a single-parameter class, but carrying the @a@ around
+-- lets us give a better error message when @'node'@ is called instead
+-- of @'nodeKeyed'@.
+--
+class GenEntityKey (s :: KeySourceType) a where
+  type KeyConstraint s a :: Constraint
```
eborden: This is pretty interesting. I don't think I've seen `Constraint` associated type families :eyes:

graphula-core/src/Graphula.hs
```diff
@@ -429,85 +467,234 @@ class HasDependencies a where
...
-nodeWith
-  :: forall a m . GraphulaContext m '[a] => Dependencies a -> m (Entity a)
-nodeWith = flip nodeEditWith id
-
-{-|
-  Generate and edit a value that does not have any dependencies.
+nodeKeyed
+  :: forall a m
+   . GraphulaContext m '[a]
+  => Key a
```
eborden: If I generate a `SourceDefault` or `SourceArbitrary` this will not throw a type error?

graphula-core/src/Graphula.hs
cdparks: Lol, I think you've forgotten that _you_ added

```haskell
class MonaGraphulaBackend m where
  type Logging m :: * -> Constraint
  type Generate m :: * -> Constraint
```

graphula-core/src/Graphula.hs
cdparks: Yep

```plaintext
/home/ubuntu/code/graphula/graphula-core/test/README.lhs:207:10: error:
    • Cannot generate a value of type ‘E’ using ‘node’ since
            
        instance HasDependencies E where
          type KeySource E = 'SourceExternal
            
      Possible fixes include:
      • Use ‘nodeKeyed’ instead of ‘node’
      • Change ‘KeySource E’ to 'SourceDefault or 'SourceArbitrary
    • In a stmt of a 'do' block: _ <- node @E () mempty
      In the second argument of ‘($)’, namely
        ‘do Entity aId _ <- node @A () mempty
            liftIO $ putStrLn "A"
            Entity bId b <- node @B (only aId) mempty
            liftIO $ putStrLn "B"
            ....’
      In the expression:
        runGraphulaT runDB
          $ do Entity aId _ <- node @A () mempty
               liftIO $ putStrLn "A"
               Entity bId b <- node @B (only aId) mempty
               liftIO $ putStrLn "B"
               ....
    |       
207 |     _ <- node @E () mempty
    |          ^^^^^^^^^^^^^^^^^
```

That's what the big mess of `TypeError` constraints is doing up above

graphula-core/src/Graphula.hs
eborden: So I did!

`i-have-no-memory-of-this-place.gif`

graphula-core/src/Graphula.hs
eborden: Beautiful

graphula-core/src/Graphula.hs
cdparks: Yes, I currently allow you to specify a key for those types, e.g.:

```haskell
instance HasDependencies D where
  type KeySource D = 'SourceArbitrary

deriving instance {-# OVERLAPPING #-} Arbitrary (Key D)

...
  k <- liftIO $ generate arbitrary
  _ <- nodeKeyed @D k () mempty
```

^^^ that one is probably harmless. This is probably a mistake:

```haskell
instance HasDependencies A where
  KeySourceA = 'SourceDefault

let k = AKey 1
_ <- nodeKeyed @A k () mempty
```

Tho, worst case scenario, you get a gen-failure when you re-run the test.

It'd be easy to disallow both cases:

```haskell
nodeKeyed
  :: forall a m
   . ( GraphulaContext m '[a]
     , KeySource a ~ 'SourceExternal
     )
  => Dependencies a
  -> NodeOptions a
  -> (m (Entity a))
```

If we just want to disallow the `'SourceExternal` case but allow `'SourceArbitrary`, it's a little more work.

cdparks: Mm, even the `'SourceArbitrary` case is only safe for `nodeKeyed` if your `Arbitrary` instance always returns a unique key (e.g. `UUID`), so maybe we should disallow both.

eborden: I don't think there is anything in principle with allowing `SourceArbitrary` and `SourceDefault` from using `keyed`. My open question is whether those uses are more likely to be intentional or errors. :man_shrugging: I think I'm fine with allowing it and seeing how it shakes out.

cdparks: Yep, same

cdparks:
Ok, before I merge this one last pass on the final API.

### Proposed API

```haskell
node      ::          Dependencies a -> NodeOptions a -> (m (Entity a))
nodeKeyed :: Key a -> Dependencies a -> NodeOptions a -> (m (Entity a))
```

This is what I have, and I think it's fine.

### Don't prefix `nodeKeyed`

```haskell
node  ::          Dependencies a -> NodeOptions a -> (m (Entity a))
keyed :: Key a -> Dependencies a -> NodeOptions a -> (m (Entity a))
```

I don't feel strongly about this one. I can take it or leave it.

### Shorthand for root nodes

```haskell
root      :: Dependencies a ~ () =>                            NodeOptions a -> (m (Entity a))
node      ::                                 Dependencies a -> NodeOptions a -> (m (Entity a))
rootKeyed :: Dependencies a ~ () => Key a ->                   NodeOptions a -> (m (Entity a))
nodeKeyed ::                        Key a -> Dependencies a -> NodeOptions a -> (m (Entity a))
```

This only saves a couple keystrokes, and we _have_ to prefix `node/rootKeyed`. I don't like it, but mentioned for completeness since the shipped API does have shorthand for roots.

```haskell
node () mempty
root mempty
nodeKey key () mempty
rootKey key mempty
```

eborden:
I have no afinity for shortening `nodeKeyed` to `keyed`. As well I don't think we need to add `root` variants to the API, they don't really pass the [fairbarn threshold](https://wiki.haskell.org/Fairbairn_threshold).

PR #35, opened 2020-03-11
GenEntityKey -> Internal.GenerateKeyInternal, expose GenerateKey subclass

cdparks:
This class is not implemented by clients, but it can appear in highly polymorphic type signatures.

For example, in `fancy-api`, there's some generic code for setting up a session with some questions:

```haskell
genSessionWithQuestions
  :: forall assignment session answer question m
   . ( MonadIO m
     , AssignmentFamily assignment session answer
     , HasFieldCompletedAt session
     , HasFieldQuestionId answer
     , ForeignQuestion answer ~ question
     , Dependencies answer ~ (Key session, Key question)
     , GraphulaContext m '[assignment, assignment, session, question, answer]
     , KeySource assignment ~ 'SourceDefault
     , KeySource session ~ 'SourceDefault
     , KeySource answer ~ 'SourceDefault
     , KeySource question ~ 'SourceArbitrary
     )
  => Dependencies assignment
  -> (Key assignment -> Dependencies session)
  -> Dependencies question -- ^ Set up question 1
  -> Dependencies question -- ^ Set up question 2
  -> (answer -> answer) -- ^ Edit answer 1
  -> (answer -> answer) -- ^ Edit answer 2
  -> m (Entity assignment, Entity session)
```

Using `KeySource _` is actually overly constrained. We'd prefer

```haskell
   . ( MonadIO m
      ...
     , GenerateKey assignment
     , GenerateKey session
     , GenerateKey answer
     , GenerateKey question
     )
  => ...
```

so that the choice of `KeySource` isn't fixed until later. For a client to be able to write this, we need to expose `GenerateKey`, which looks like

```haskell
class
  (GenerateKeyInternal (KeySource a) a, KeyConstraint (KeySource a) a)
  => GenerateKey a

instance
  (GenerateKeyInternal (KeySource a) a, KeyConstraint (KeySource a) a)
  => GenerateKey a
```

`GenerateKeyInternal` is available in `Graphula.Internal` and actually implements the key generation that triggers off of a type's `KeySource`.


graphula-core/src/Graphula.hs
```diff
@@ -56,6 +56,10 @@ module Graphula
     -- * Declaring Dependencies and key source
   , HasDependencies(..)
   , KeySourceType(..)
+    -- Not implemented by clients, but may appear
```
eborden: Oh actually can you make these haddocks

cdparks: Yeh

cdparks:
Description updated, chatted with @eborden about subclass approach.

PR #37, opened 2020-05-21
Don't require JSON instances on node types

pbrisbin:
Instead, require Generic and dump/load JSON via Generic-based instances
internally via an AsJSON newtype.


cdparks:
The `D` test entity uses `UUIDKey` which is a `newtype` over `UUID` that doesn't implement `Generic` so if it builds, it must work. Also, can you remove the `test.db` file? That can probably go in `.gitignore`

pbrisbin:
It looks like we don't have CI here (I'll drop an Issue), and I'm going to skip the overall Restyle of the bench file since I only made a minor change to it.

cdparks:
From Slack:

**Chris**: i figured it out. we don't log the whole `Entity X`, just `X`, so the README isn't testing this
**Pat**: that's convenient
**Chris**:  but, if you add something to the columns that doesn't have `ToJSON`, then you blow up
```plaintext
/home/ubuntu/code/graphula/graphula-core/test/README.lhs:188:21: error:
    • No instance for (aeson-1.4.2.0:Data.Aeson.Types.ToJSON.ToJSON
                         GUID)
        arising from a use of 'node'
    • In a stmt of a 'do' block: Entity dId _ <- node @D () mempty
      In the second argument of '($)', namely
        'do Entity aId _ <- node @A () mempty
            liftIO $ putStrLn "A"
            Entity bId b <- node @B (only aId) mempty
            liftIO $ putStrLn "B"
            ....'
      In the expression:
        runGraphulaT runDB . runGraphulaLoggedWithFileT "x.graphula"
          $ do Entity aId _ <- node @A () mempty
               liftIO $ putStrLn "A"
               Entity bId b <- node @B (only aId) mempty
               liftIO $ putStrLn "B"
               ....
    |
188 |     Entity dId _ <- node @D () mempty
    |                     ^^^^^^^^^^^^^^^^^
```
**Chris**: so `AsJSON X` allows `X` to be serialized if all of its fields have `ToJSON`
 **Chris**: but `X` itself doesn't have to have `ToJSON`

PR #40, opened 2020-05-21
House cleaning

pbrisbin:
- [x] Git ignore test.db
- [x] Stack lockfile
- [x] Circle config
- [x] Update to latest LTS
- [x] Restyle all files
- [x] Make CI happy


graphula-core/src/Graphula.hs
```diff
@@ -617,7 +612,7 @@ attempt maxEdits maxInserts source = loop 0 0
 
 -- | For entities that only have singular 'Dependencies'. It uses data instead
 -- of newtype to match laziness of builtin tuples.
```
pbrisbin, line 614: Just noticed this comment. I guess I shouldn't blindly follow HLint? This laziness point is important?

cdparks: Eh, my opinion on this has evolved. Now I think a singleton tuple should be representationally identical to its parameter. Technically a breaking change, but no one should be relying on the indirection here. We can do a version bump once you change the exception types tho.

graphula-core/test/README.lhs
```diff
@@ -18,6 +18,7 @@ Graphula is a simple interface for generating persistent data and linking its de
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
```
cdparks, line 21: Persistent!

graphula-core/src/Graphula.hs
pbrisbin: Cool, I'll nuke the comment before merging.

PR #41, opened 2020-05-22
Remove Replay so we can remove JSON instance requirement

pbrisbin:
Changes:

- ff5e86f Remove Replay functionality

  Requiring that node types have JSON instances is problematic for us for
  nuanced, domain-specific reasons. By removing replay functionality, which we
  don't use, we believe we can then remove the JSON requirement by logging
  graphs, which we do use, as a serialization format that need not support
  de-serialization (e.g. `Show`).

  Ideally, this would be more directly configurable by end-users, such that
  anyone who wanted Replay could accept the JSON requirement, without forcing it
  on those that don't. But that is deferred for now since we are the only end
  users, as far as we know. (And it seemed mad hard yo.)

- 333f504 Use `Show`, not `ToJSON` in logging graph nodes

  **Concern**: we only write the graph to on failures, so performance at that
  step is probably not a concern, but we maintain the graph in the `IORef (Seq
  x)` all the time (AFAICT), so this will change what was `Seq Value` to `Seq
  Text` in an area that is exercised every test. That coo?

  I'm happy to use anything other than `Show` here. The requirements are:

  - Human readable
  - Isn't cumbersome for node types to implement
  - Isn't JSON ;)

  I picked `Show` since it's the least cumbersome thing for us to move to when
  this lands, as ~all our entities already have that instance. I'm also down to
  bring in (e.g.) `groom` to make the `Show`n values even more prettier, if we
  want.

### DIFF
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -73,8 +73,6 @@ module Graphula
   , runGraphulaLoggedT
   , runGraphulaLoggedWithFileT
   , GraphulaLoggedT
-  , runGraphulaReplayT
-  , GraphulaReplayT
     -- ** Frontends
   , runGraphulaIdempotentT
   , GraphulaIdempotentT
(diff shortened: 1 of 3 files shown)


graphula-core/src/Graphula.hs
```diff
@@ -279,26 +267,26 @@ runGraphulaIdempotentT action = mask $ \unmasked -> do
...
 
 instance MonadIO m => MonadGraphulaBackend (GraphulaLoggedT m) where
-  type Logging (GraphulaLoggedT m) = ToJSON
+  type Logging (GraphulaLoggedT m) = Show
   type Generate (GraphulaLoggedT m) = Arbitrary
   generateNode = liftIO $ generate arbitrary
   logNode n = do
     graphLog <- ask
-    liftIO $ modifyIORef' graphLog (|> toJSON n)
+    liftIO $ modifyIORef' graphLog (|> pack (show n))
```
cdparks, line 282: `Seq` is lazy in its elements so we're not really running `pack . show` until (or unless) the sequence is written to a file.

pbrisbin:
@eborden that sounds like a great solution! I know that `hspec` supports `--seed`, which impacts any QuickCheck properties, I'm just not sure if any and all uses of `Arbitrary` within the suite will inherit that. Should be (relatively) easy to test by comparing the graph dumps of a second run with the same seed. That's a bit of game-changer for investigating unreliable tests IMO, if it works.

eborden:
I'm going to open an issue for that. There are some complexities around concurrency, utilizing the seed, etc, that are unknown. Likely `GraphulaT` would need to become a `State` monad to accept and track a PRNG instead of utilizing `IO`.

Issue #42, opened 2020-05-26
Allow seeded random generation


eborden:
Replay was removed here: https://github.com/freckle/graphula/pull/41

The implementation of replay was never ideal. It required a lot of moving parts and the integration of graphula into our codebase actually makes it hard to utilize.

A better path for replay is probably leaning harder on the PRNG that underlies generation. Instead of logging/loading everything, we should be able to thread the seed for this tests PRNG to create replay-ability. This is likely already possible via hspec or other testing frameworks.

There are some complexities around concurrency, utilizing the seed, etc, that are unknown. Likely `GraphulaT` would need to become a `State` monad to accept and track a PRNG instead of utilizing IO.

cdparks:
I'll have to think about how this interacts with ids that we get back from the DB, since those can't be traced back to the random seed.

eborden:
@cdparks I think IDs you get back from the DB can be thought of as transparent references. They shouldn't interact in any meaningful way with deterministic generation, since the generation is unconcerned with their value.

eborden:
Randomly generated IDs however could interact in a fun way. They'll be deterministic-ally generated, so replay will require clearing entities out of the DB.

cdparks:
I meant more when we use the `Arbitrary` mechanism outside the usual `node` calls:

```haskell

    it "should fetch a null set of unmanaged courses" . withGraph runDB $ do
      -- Uses our generate, and thus our QCGen
      schoolAdmin <- simpleSchoolAdmin
      -- Uses QC's generate, and thus QC's QCGen
      sisId <- liftIO $ generate arbitrary
      -- Back to us
      school <- genAdministeredSchoolWith schoolAdmin $ setField SchoolSisId $ Just sisId
```

pbrisbin:
The export of `Graphula.Arbitrary.generate` solves that now right? If you use that function for these, those would also be deterministic by the passed seed.

cdparks:
Yeah, we'll just need to make sure we're hiding the `Arbitrary` one in our tests.

PR #43, opened 2020-06-01
Utilize stack local random generator

eborden:
Log based replay was removed from `graphula`. Log based replay was
necessary because random generation utilized non-deterministic `IO`.
Graphula can have deterministic random generation. Deterministic
generation allows seeds to be utilized for replay instead of logs.

- Deterministic generation requires holding onto a `QCGen` instance as
  state. This state is read, split and updated with every `generate` call.
- The `ReaderT` `IORef` pattern is used to allow `MonadUnliftIO`.
- The `Generate` associate type has been removed in favor of direct use
  of `Arbitrary`. `Arbitrary` has been used directly in graphula for a
  while now, via key generation. There is no reason to deny this fact
  with the complexity of an associated type.
- `withGraphulaT` now accepts an optional `seed`. Without the seed it
  uses `IO` to produce a random seed.
- On failure the seed is printed with the hunit failure.

This is in draft stage. If possible I'd like to figure out how to incorporate
`hspec`'s seed handling, which may mean the addition of a `graphula-hspec`
package. Though this is usable in its current state.

### DIFF
--- a/graphula-core/src/Graphula.hs
+++ b/graphula-core/src/Graphula.hs
@@ -87,10 +87,10 @@ import Prelude hiding (readFile)
 
 import Control.Monad (guard, (<=<))
 import Control.Monad.IO.Unlift
-import Control.Monad.Reader (MonadReader, ReaderT, ask, runReaderT)
+import Control.Monad.Reader (MonadReader, ReaderT, ask, asks, runReaderT)
 import Control.Monad.Trans (MonadTrans, lift)
 import Data.Foldable (for_, traverse_)
-import Data.IORef (IORef, modifyIORef', newIORef, readIORef)
+import Data.IORef (IORef, modifyIORef', newIORef, readIORef, writeIORef)
 import Data.Kind (Constraint, Type)
 import Data.Proxy (Proxy(..))
 import Data.Semigroup.Generic (gmappend, gmempty)
@@ -119,9 +119,12 @@ import Graphula.Internal
 import System.Directory (createDirectoryIfMissing, getTemporaryDirectory)
 import System.IO (Handle, IOMode(..), hClose, openFile)
 import System.IO.Temp (openTempFile)
+import System.Random (randomIO, split)
 import Test.HUnit.Lang
   (FailureReason(..), HUnitFailure(..), formatFailureReason)
-import Test.QuickCheck (Arbitrary(..), Gen, generate)
+import Test.QuickCheck (Arbitrary(..), Gen)
+import Test.QuickCheck.Gen (unGen)
+import Test.QuickCheck.Random (QCGen, mkQCGen)
 import UnliftIO.Exception
   (Exception, SomeException, bracket, catch, mask, throwIO)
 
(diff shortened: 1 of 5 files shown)


graphula-core/src/Graphula.hs
```diff
@@ -554,3 +561,13 @@ newtype Only a = Only { fromOnly :: a }
 
 only :: a -> Only a
 only = Only
+
+generate :: (MonadIO m, MonadGraphulaBackend m) => Gen a -> m a
```
eborden, line 565: I'm thinking about renaming this and exposing it. That way anything in tests that utilizes random generation can use the `graphula` seed and remain deterministic, instead of falling back to `IO`.

eborden:
This addresses https://github.com/freckle/graphula/issues/42

eborden:
Oh, random side note. This should also improve performance in the presence of concurrency. `IO` based random generation utilizes the global generator, which must be locked, accessed, and updated on every generation. By switching to a local `PRNG` we remove a synchronization point between all threads. Likely `graphula` isn't high performance enough to benefit in any meaningful way, but it is there.

graphula-core/src/Graphula.hs
pbrisbin: I kind of wish we could export it at this name, to make it clear(er) that you _should_ do that to keep it all deterministic within the test. But I also agree with the fear of import churn to deal with sudden naming conflicts in any existing tests also using `Arbitrary`.

What if we exported it as `generate`, but from a new module named `Graphula.Arbitrary`? Not super discoverable, but at least a little clearer (IMO) that, if you want to do `Arbitrary` stuff in a `Graphula` test, you should use these (conflicting) names from that module instead?

eborden:
@pbrisbin utilizing `--seed` is my hope. I've dug a bit and it seems it will require creating an instance of `Example`, but I haven't gotten much farther than that.

pbrisbin:
> it seems it will require creating an instance of `Example`

From my prior experience dealing with `Example`, this little tidbit tells me that it's both _possible_ and annoying-AF to do what we want :joy: But we'll see!

eborden:
Running this in our tests:
image

graphula-core/src/Graphula/Arbitrary.hs
```diff
@@ -0,0 +1,34 @@
...
+import Test.QuickCheck (Gen)
+import Test.QuickCheck.Gen (unGen)
+
+-- | Run a generator
+--
+-- This is akin to 'Test.QuickCheck.generate', but utilizing graphula's
+-- generation. The size passed to the generator is always 30; if you want
+-- another size then you should explicitly use 'Test.QuickCheck.resize'.
+--
+generate :: (MonadIO m, MonadGraphulaBackend m) => Gen a -> m a
```
cdparks, line 26: Neat. We'll need to remember to replace the ad-hoc `liftIO $ generate arbitrary`'s we have in our tests for, e.g. `UUID`s and emails.

eborden:
Merging this current version. I'm still exploring the options of a `graphula-hspec` utilizing `Example`, but no need to block this bit.

Issue #45, opened 2020-08-25
include cabal files since stack is deprecating hpack only for dependencies


codygman:
https://github.com/commercialhaskell/stack/issues/5210

```
DEPRECATED: The package at Repo from https://github.com/freckle/graphula.git, commit c20e88b8de768728dff246b458fde55068779322 in subdir graphula-core does not include a cabal file.
Instead, it includes an hpack package.yaml file for generating a cabal file.
This usage is deprecated; please see https://github.com/commercialhaskell/stack/issues/5210.
Support for this workflow will be removed in the future.

DEPRECATED: The package at Repo from https://github.com/freckle/graphula.git, commit c20e88b8de768728dff246b458fde55068779322 in subdir persistent-arbitrary does not include a cabal file.
Instead, it includes an hpack package.yaml file for generating a cabal file.
This usage is deprecated; please see https://github.com/commercialhaskell/stack/issues/5210.
Support for this workflow will be removed in the future.
```

eborden:
Thanks for the poke. I'm curious @codygman are you using `graphula` in a project?

codygman:
@eborden I'm testing it out at least :smile: It seems to be exactly what our team needs to ease generating test data.

eborden:
@codygman if you need a hand let us know. This tool has been very effective for our team, but generative data has its own fun edge cases that we've learned to manage through trial and error. Any hard edges you hit likely have solutions or might allow us to push fixes into `graphula` proper.

codygman:
@eborden Thanks a ton for the offer and quick turn around time! I'll definitely try to note any hard edges, hopefully if I encounter we can make improvements.

PR #46, opened 2020-08-25
Remove cabal file from gitignore

eborden:
Addresses: https://github.com/freckle/graphula/issues/45

https://github.com/commercialhaskell/stack/issues/5210

The stack team is now recommending that projects store their cabal file
instead of utilizing just `hpack` generation.


eborden:
Yeah, I'll open another PR for that.

PR #48, opened 2020-11-17
Re-organize and update project

pbrisbin:
Removes the `persistent-arbitrary` package, because we've been creating that instance client side ourselves, and can do so in our test suite too. At that point, the `graphula-core` sub-package made little sense, so I moved all the code to top-level. Then I updated the Stackage resolver, added our usual `-Weverything`, and addressed all the warnings.


src/Graphula/Internal.hs
```diff
@@ -8,6 +8,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-missing-export-lists #-}
```
pbrisbin, line 11: I'll address this before merging.

test/README.lhs
```diff
@@ -19,15 +19,17 @@ Graphula is a simple interface for generating persistent data and linking its de
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE UndecidableInstances #-}
-module Main where
+{-# OPTIONS_GHC -fno-warn-missing-deriving-strategies #-}
```
pbrisbin, line 22: Necessary for the persistent quasi-quoter.

src/Graphula.hs
```diff
@@ -43,6 +25,23 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
 
+-- |
+--
+-- Graphula is a compact interface for generating data and linking its
+-- dependencies. You can use this interface to generate fixtures for automated
+-- testing.
+--
+-- The interface is extensible and supports pluggable front-ends.
```
eborden, line 34: Is the pluggable frontends still true?

eborden (approved):

cdparks (approved):

eborden:
Looks like you merged this to `master` instead of `main`.

pbrisbin:
Doh. What can we do to make that error and not just create the branch?

eborden:
Right, gotta delete the branch.

PR #49, opened 2020-11-17
Prepare release

pbrisbin:


cdparks (approved):
I guess that last PR was technically non-breaking?

pbrisbin:
I didn't give it any thought. Since this will be the first version actually on Hackage I didn't think it mattered.

PR #50, opened 2020-11-18
Rename package to just graphula

pbrisbin:
There is no `graphula-{not core}` anymore, so the `-core` is superfluous. I had
intended to upload this package to Hackage as graphula, but missed this little
`name`.

I assume I can just re-release like this to create `graphula` on Hackage too,
but what do I do with the `graphula-core` there now?

### DIFF
--- a/graphula.cabal
+++ b/graphula.cabal
@@ -4,9 +4,9 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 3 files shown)


eborden (approved):

eborden:
I believe you can deprecate a package...

eborden:
On the maintainers page

http://hackage.haskell.org/package/graphula-core/maintain

image

pbrisbin:
Perfect, will do.

PR #51, opened 2020-12-07
Stackage nightly ghc810 updates

codygman:
I didn't actually mean to open this PR yet, but since I did:

I can clean this up but wanted to see what your thoughts were on it.

### DIFF
--- a/src/Graphula.hs
+++ b/src/Graphula.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE DeriveGeneric #-}
@@ -168,12 +169,15 @@ instance MonadTrans (GraphulaT n) where
   lift = GraphulaT . lift
 
 instance MonadUnliftIO m => MonadUnliftIO (GraphulaT n m) where
-  {-# INLINE askUnliftIO #-}
-  askUnliftIO = GraphulaT $ withUnliftIO $ \u ->
-    return $ UnliftIO $ unliftIO u . runGraphulaT'
+#if MIN_VERSION_unliftio_core(0, 2, 0)
   {-# INLINE withRunInIO #-}
   withRunInIO inner =
     GraphulaT $ withRunInIO $ \run -> inner $ run . runGraphulaT'
+#else
+  {-# INLINE askUnliftIO #-}
+  askUnliftIO = GraphulaT $ withUnliftIO $ \u ->
+    return $ UnliftIO $ unliftIO u . runGraphulaT'
+#endif
 
 instance MonadIO m => MonadGraphulaBackend (GraphulaT n m) where
   type Logging (GraphulaT n m) = NoConstraint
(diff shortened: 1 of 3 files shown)


src/Graphula.hs
```diff
@@ -168,12 +169,15 @@ instance MonadTrans (GraphulaT n) where
   lift = GraphulaT . lift
 
 instance MonadUnliftIO m => MonadUnliftIO (GraphulaT n m) where
-  {-# INLINE askUnliftIO #-}
-  askUnliftIO = GraphulaT $ withUnliftIO $ \u ->
-    return $ UnliftIO $ unliftIO u . runGraphulaT'
+#if MIN_VERSION_unliftio_core(0, 2, 0)
```
eborden, line 172: I would love if we could avoid CPP, but this `unliftio` transition might make it impossible for the time being.

stack.yaml
```diff
@@ -1,7 +1,10 @@
-resolver: lts-16.22
+resolver: nightly-2020-12-03
```
eborden, line 1: I'm not sure what our stance is on nightly resolvers. @pbrisbin do you have a policy for Freckle's OSS projects and snapshot alignment?

src/Graphula.hs
tfausak: It's too bad that supporting both versions requires CPP 😞

pbrisbin (commented):
Hi there-

Thanks very much for this contribution. We don't have much in the way of "policy" for OSS, but I can comment on the things going on here from my perspective, which should mostly align with the team.

**Nix**

I'm happy to keep anything nix-related in `stack.yaml` as long as it doesn't impact non-Nix users (I assume it won't), so :+1:.

**Nightly**

I do want to test projects on nightly on CI. That's pretty much required for us to be in Stackage, which we're not currently but aim to be generally. However, I wouldn't change the default `stack.yaml` to nightly, I'd like to leave that as "latest LTS" always. This prevents any transient compatibility issues from biting new contributors.

What we typically do is:

- Add the [`build-test-nightly`](https://circleci.com/developer/orbs/orb/pbrisbin/stack-build#jobs-build-test-nightly) Job from our stack-build Orb to CI, on push and timed each night
- If that works with `stack.yaml` great. If not, add `stack-nightly.yaml` specifically for it

[Here's an example](https://github.com/pbrisbin/bugsnag-haskell/blob/master/.circleci/config.yml#L44-L65) from one of my non-Freckle OSS projects.

**CPP / unliftio**

First, we typically set up projects to build on CI across latest GHC and 1-3 versions back depending on how difficult it is. [Here's an example](https://github.com/freckle/bcp47/blob/master/.circleci/config.yml).

The choice of how many versions of GHC we support and build on usually comes down to CPP. Our OSS stuff isn't widely-used so we don't really take on CPP as a way to support a wide range of GHCs.

I look at unliftio similarly. Therefore, I would remove the CPP and add a minimum bound on unliftio. We can probably still build across a few GHCs by bringing in the newer unliftio as an `extra-dep` in those specific `stack-x.yaml` configurations.

If that *doesn't* go well (i.e. if we can't target a few resolvers/GHCs without bringing in a larger tree of extra-deps), then I'd be more inclined to take on the CPP. Or if anyone on the team feels it's valuable.

So-- if that all gels with the team --I think this PR could be merged if we removed the resolver change and add an unliftio minimum bound. I would update CI to exercise nightly (and/or build on other GHCs) in a separate PR, which I'd probably wait for before releasing a new version.

stack.yaml
pbrisbin: Mentioned in my top-level review: I like to keep the default `stack.yaml` on "latest LTS" but then also test with `nightly` on CI.

pbrisbin:
Heads-up: I did some of this work in #53

src/Graphula.hs
pbrisbin: In older versions, `askUnliftIO` was an optional member of the type class anyway, so just removing it should work across them.

tfausak:
This was made obsolete by #53. Thanks @pbrisbin!

pbrisbin:
I didn't realize I had subsumed almost all of this PR there -- so I'm going to close this. The only thing I didn't add was the Nix bits, so feel free to open a new PR of just that.

PR #53, opened 2020-12-22
Build on nightly

pbrisbin:
~~Newer GHC adds two warnings which we can only disable on nightly, since they're
unknown options to older GHCs. Therefore, they were added directly on the CLI in
CI.~~

~~Newer GHC (or Cabal?) also adds the unnecessary package warning, which I think
is very valuable so we can't disable it to support the `markdown-unlit`
executable dependency required for the `readme` spec. To avoid that, we have to
build it explicitly on CI. Local development will also hit an error unless it's
installed explicitly there. We should investigate how to smooth this -- e.g.
bring back `make setup`?~~

Remove `askUnliftIO`, as it's no longer part of the type class in newer
`unliftio-core`. **NOTE**: this is not a compatibility issue since that class
member was always optional anyway.

Unused type variables are now also flagged. Replacing them with underscores
solves this, but *just* an `_` is a syntax error in older GHCs, so we have to
use `_a`, even though the "a" part adds no information.


tfausak:
> Newer GHC adds two warnings which we can only disable on nightly, since they're unknown options to older GHCs.

I don't remember how to do this with hpack, but you can conditionally enable warnings in Cabal files like this:

``` yaml
ghc-options: -Wfoo
if impl(ghc >= 8.10)
  ghc-options: -Wbar
```

> Newer GHC (or Cabal?) also adds the unnecessary package warning, which I think is very valuable so we can't disable it to support the markdown-unlit executable dependency required for the readme spec. 

For what it's worth this is a new GHC feature: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/154

You may be able to trick GHC into thinking that `markdown-unlit` is used by added an unnecessary import of it:

``` hs
import Text.Markdown.Unlit ()
```

pbrisbin:
Thanks @tfausak I'll try to incorporate both suggestions!

cdparks (approved):

PR #56, opened 2021-02-09
Project updates

pbrisbin:
- Replace Circle with GitHub Actions
- Relax dependency bounds

This ensures --pvp-bounds=both doesn't exclude the latest packages on
Hackage (or Stackage nightly).

https://packdeps.haskellers.com/feed?needle=graphula

- Version bump


.github/workflows/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.stack
+            ./.stack-work
+          key: ${{ runner.os }}-${{ hashFiles('stack.yaml') }}-${{ hashFiles('*.cabal') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ hashFiles('stack.yaml') }}-
+            ${{ runner.os }}-
```
eborden, line 19: This seems to be getting repeated. Should we have a `freckle/stack-cache-action`?

eborden (approved):

.github/workflows/test.yaml
pbrisbin: I thought that from the very first project I typed it; I even started on one, but it's non trivial and I gave up. Feel free to check out the [npm library](https://github.com/actions/toolkit/tree/main/packages/cache), you would need to use to accomplish it.

PR #61, opened 2021-05-05
Support persistent-2.13

parsonsmatt:


eborden (approved):
LGTM, thanks for the bump!

pbrisbin:
I'll merge and release this today; thanks!

parsonsmatt:
`persistent-template-2.12` is a dummy package that has docs that tell you to just use `persistent` from 2.12 on. It doesn't export anything.

pbrisbin:
So all the more reason to remove it from the dependencies and fix this error?

```
 <no location info>: error: [-Wunused-packages, -Werror=unused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - persistent-template-2.12.0.0
```

Or am I missing something?

parsonsmatt:
What's really annoying is that you'd want to conditionally depend on it. If `persistent < 2.12`, then you need `persistent-template`. But if `persistent >= 2.12`, then you don't need it. And IIRC cabal does not have a way to specify that.

pbrisbin:
Sigh. And there's no way to disable `-Wunused-packages` except globally, is there?

pbrisbin:
I just did a release that removed upper bounds (we've decided the maintenance burden is too much for us and are going to try being more lax for now) to unblock https://github.com/commercialhaskell/stackage/issues/6010.

I didn't add the extra-deps, so the `-template` issue wasn't triggered and CI was green. Sorry for the conflicts.

parsonsmatt:
Thanks! That works for me 😄

PR #62, opened 2021-05-06
Bump persistent bounds

5outh:
graphula works fine with the new version of persistent, so I'm just relaxing the bounds.


5outh:
Wooops, jinx with #61 , closing!

PR #66, opened 2022-04-14
Support `persistent-2.14`

### LINKED ISSUE #65: `persistent-2.14` support

parsonsmatt:
This PR provides support for `persistent-2.14` by CPPing in a `GraphulaSafeToInsert` type that is either an alias for `SafeToInsert` or `const () :: Constraint`.

Fixes #65

### DIFF
--- a/src/Graphula/Class.hs
+++ b/src/Graphula/Class.hs
@@ -25,7 +25,7 @@ import Database.Persist (Entity(..), Key, PersistEntity, PersistEntityBackend)
 import Database.Persist.Sql (SqlBackend)
 import Test.QuickCheck.Random (QCGen)
 #if MIN_VERSION_persistent(2,14,0)
-import Database.Persist.Class.PersistEntity
+import Database.Persist.Class.PersistEntity (SafeToInsert)
 #endif
 
 -- | A class that provides backwards compatibility with @persistent-2.14@
--- a/src/Graphula/Class.hs
+++ b/src/Graphula/Class.hs
@@ -28,9 +28,11 @@ import Test.QuickCheck.Random (QCGen)
 import Database.Persist.Class.PersistEntity
 #endif
 
--- | A clas that provides backwards compatibility with @persistent-2.14@.
+-- | A class that provides backwards compatibility with @persistent-2.14@
+--
 -- If you are using that version or above, then this is a class alias for
--- @SafeToInsert@. Otherwise, it is an identity alias.
+-- 'SafeToInsert'. Otherwise, it is an identity alias.
+--
 type GraphulaSafeToInsert a =
 #if MIN_VERSION_persistent(2,14,0)
     SafeToInsert a
--- a/src/Graphula/Class.hs
+++ b/src/Graphula/Class.hs
@@ -33,12 +33,21 @@ import Database.Persist.Class.PersistEntity
 -- If you are using that version or above, then this is a class alias for
 -- 'SafeToInsert'. Otherwise, it is an identity alias.
 --
-type GraphulaSafeToInsert a =
+class
 #if MIN_VERSION_persistent(2,14,0)
-    SafeToInsert a
+    (SafeToInsert a) =>
 #else
-    () :: Constraint
+    () =>
 #endif
+    GraphulaSafeToInsert a
+
+instance
+#if MIN_VERSION_persistent(2,14,0)
+    (SafeToInsert a) =>
+#else
+    () =>
+#endif
+    GraphulaSafeToInsert a
 
 type MonadGraphula m
   = (Monad m, MonadIO m, MonadGraphulaBackend m, MonadGraphulaFrontend m)
--- a/src/Graphula/Class.hs
+++ b/src/Graphula/Class.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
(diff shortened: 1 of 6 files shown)


stack.yaml
```diff
@@ -2,3 +2,6 @@ resolver: lts-19.1
 
 ghc-options:
   "$locals": -fwrite-ide-info
+
+extra-deps:
+  - persistent-2.14.0.0
```
parsonsmatt, line 7: this is how i tested, can definitely back out

src/Graphula/Class.hs
```diff
@@ -22,13 +24,26 @@ import Data.Kind (Constraint, Type)
 import Database.Persist (Entity(..), Key, PersistEntity, PersistEntityBackend)
 import Database.Persist.Sql (SqlBackend)
 import Test.QuickCheck.Random (QCGen)
+#if MIN_VERSION_persistent(2,14,0)
+import Database.Persist.Class.PersistEntity
+#endif
+
+-- | A clas that provides backwards compatibility with @persistent-2.14@.
```
mjgpy3, line 31: Nit
```suggestion
-- | A class that provides backwards compatibility with @persistent-2.14@.
```

src/Graphula/Node.hs
```diff
@@ -134,6 +135,7 @@ nodeKeyed
      , PersistEntityBackend a ~ SqlBackend
      , PersistEntity a
      , Typeable a
+     , GraphulaSafeToInsert a
```
parsonsmatt, line 138: so, technically, this is a change to the signature of the function, which is a breaking change

but also technically, it's only a breaking change if `persistent` itself is updated to `2.14`. Otherwise this evaluates out to a `()` and should discharge

I can also do another approach where I CPP this in at each use-site, which will prevent the API from changing for users unless they are on `persistent-2.14` anyway.

mjgpy3 (approved):
Looks good to me! I'm gonna bring in @pbrisbin for the stack.yaml question.

stack.yaml
pbrisbin: IIUC, `persistent-2.14` has landed in `nightly` (we got a notification about it breaking us), so the `stack-nightly.yaml` file should be a good test (and is run on CI).

src/Graphula/Node.hs
pbrisbin: I'm personally OK with us only doing a minor bump ourselves given this nuance. Curious what other Frecklers thing.

I don't think peppering the CPP around is worthwhile. I think we can either bend the rules or just do a major bump.

src/Graphula/Class.hs
```diff
@@ -22,13 +24,26 @@ import Data.Kind (Constraint, Type)
 import Database.Persist (Entity(..), Key, PersistEntity, PersistEntityBackend)
 import Database.Persist.Sql (SqlBackend)
 import Test.QuickCheck.Random (QCGen)
+#if MIN_VERSION_persistent(2,14,0)
+import Database.Persist.Class.PersistEntity
+#endif
+
+-- | A class that provides backwards compatibility with @persistent-2.14@.
+-- If you are using that version or above, then this is a class alias for
+-- @SafeToInsert@. Otherwise, it is an identity alias.
```
pbrisbin, line 33: Just rewriting this in our own style:

```suggestion
-- | A class that provides backwards compatibility with @persistent-2.14@
--
-- If you are using that version or above, then this is a class alias for
-- 'SafeToInsert'. Otherwise, it is an identity alias.
--
```

pbrisbin (commented):
If I'm correct about this being exercised via `stack-nightly.yaml`, then I would indeed back out the `stack.yaml` change.

In other projects (e.g. yesod-auth-oauth) I've actually made a special `stack-{dependency}-{version}.yaml` and added it to the matrix, if I'm looking to "test ahead". So I might consider that if I'm _not_ correct about how this is playing out in `nightly`.

Either way, not a big deal.

parsonsmatt:
Hmm, GHC 8.6.5 and below don't support `type A a = Eq a :: Constraint` syntax.

pbrisbin:
Well that's an interesting failure to only get in on resolver :thinking: is that a new (or improved) warning?

I don't mind disabling it as workaround -- I find the `README.lhs` specs can sometimes need that because the tricks needed to make the nice README can be not-great Haskell.

parsonsmatt:
It may also be a problem with how `persistent-2.14` is generating code, too.

parsonsmatt:
Anything I can do to help move this forward?

pbrisbin:
Sorry I lost track of this. I think we were waiting on you fixing the stack-nightly build-error; perhaps by disabling a warning by pragma in the `README.lhs`? The log has expired so I'm just going by my previous comment.

parsonsmatt:
I think it will be very inefficient for me to fix a CI problem since I need manual approval for each CI run. Is there an easy way for me to test that locally?

pbrisbin:
> I think it will be very inefficient for me to fix a CI problem since I need manual approval for each CI run

I thought I only had to approve a PR/contributor once. It's *every time*? Sigh.

> Is there an easy way for me to test that locally?

As far as I can tell, CI just runs `stack --resolver nightly --stack-yaml stack-nightly.yaml build --fast --pedantic --test`. Does that reproduce the error for you?

parsonsmatt:
Hmm. `brittany` fails in the presence of CPP so `restyled` isn't going to work 🤔

pbrisbin:
Restyled is not a Required Status, so all good there.

pbrisbin:
Thanks!

Issue #70, opened 2023-07-25
Cannot use with entities not `SafeToInsert`


pbrisbin:
As of lts-21, the version of persistent present needs `SafeToInsert` to use `insert`. This serves as witness that the entity in fact uses an auto-increment key and so is "safe to insert [without explicit key]". Entities where this is _not_ true must use `insertKey` and supply the key. `insertKey`, of course, does not enforce `SafeToInsert`. So safe.

In Graphula, we have only the single `insert` member in the [`MonadGraphulaFrontend`](https://hackage.haskell.org/package/graphula-2.0.2.2/docs/Graphula.html#t:MonadGraphulaFrontend) typeclass. And it always requires (our compatibility type for) `SafeToInsert`. This is because it takes a `Maybe Key` argument and uses `Persist.insert` on `Nothing` (or `Persist.insertKey` on `Just`).

Up until lts-21, this constraint was a no-op and we in fact created cases where we use Graphula on entities that _aren't_ safe to insert. Since we always supplied a `Just key` to our `insert` for these cases (based on `KeySource`), it all worked. But once that constraint becomes actually-enforced, they no longer type check.

At this point, in order to depend on newer persistent and keep up with the latest LTS, we think we need to expand the type-class to include `insertKey`, which will be used by `nodeKeyed` (not `node`), and then we can migrate these non-compiling use-cases.

This calls into question the overall design of `insert`, with its `Maybe` argument, `node` and `KeySource`, etc -- as we seem to be diverging with how persistent is tackling a similar design space. If it turns out we can't avoid a major version bump, we might consider taking larger liberties here.

However, if we _can_ do this as a minor version bump that would be preferred. It could be possible if we give `insertKey` some default. But we have to do that in a way that doesn't keep the `SafeToInsert` constraint on `nodeKeyed` when used. (For example, `insertKey = insert . Just` won't work.) :thinking: 

/cc @tomcarste

chris-martin:
Done in #73

PR #73, opened 2023-12-13
Remove inappropriate SafeToInsert constraints

chris-martin:
Persistent 2.14 introduced the `SafeToInsert` class to flag a record as being usable with the `insert`; is a record is not "safe to insert" then it must be inserted with `insertKey`.

https://github.com/freckle/graphula/pull/66 introduced a compatibility class `GraphulaSafeToInsert` to allow `graphula` to be usable with persistent >=2.14. However, it did so by placing excessive `GraphulaSafeToInsert` constraints.

- For `node` (which handles cases where the record's key source is `SourceDefault` or `SourceArbitrary`) this is somewhat overconstrained; `SafeToInsert` should be necessary when the record's key source is `SourceDefault`, but it should not be necessary when the key is `SourceArbitrary`, since we will always be generating a key to use for the insert.
- For `nodeKeyed`, the `SafeToInsert` constraint should not be needed at all, since a key has been provided.

This PR fixes those constraints, though it isn't easy to see because some of it is buried under a handful of type-level functions. (I acknowledge that this implementation could probably be made prettier, though it's "good enough for now" IMO.)

`MonadGraphulaFrontend` now needs a new `insertKeyed` function in addition to `insert`, because insertions with and without a key now require different calls to Persistent. This is the cause for the major version bump, as it will be a breaking change for any instances of the class.

Upgrading to this version of graphula (and adding `SafeToInsert` constraints to a handful of functions) is necessary and sufficient for upgrading `megarepo/backend` to the latest `persistent`.

### DIFF
--- a/src/Graphula/Dependencies.hs
+++ b/src/Graphula/Dependencies.hs
@@ -24,18 +24,29 @@ module Graphula.Dependencies
 
     -- * Non-serial keys
   , KeySourceType (..)
+  , KeySourceTypeM
+  , KeyForInsert
+  , KeyRequirementForInsert
+  , InsertWithPossiblyRequiredKey (..)
+  , Required (..)
+  , Optional (..)
   , GenerateKey
   , generateKey
   ) where
 
 import Prelude
 
-import Data.Kind (Constraint)
+import Data.Kind (Constraint, Type)
 import Data.Proxy (Proxy (..))
-import Database.Persist (Key)
+import Database.Persist (Entity (..), Key, PersistEntity, PersistEntityBackend)
+import Database.Persist.Sql (SqlBackend)
 import GHC.Generics (Generic)
 import GHC.TypeLits (ErrorMessage (..), TypeError)
 import Generics.Eot (Eot, HasEot, fromEot, toEot)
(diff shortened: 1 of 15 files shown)


src/Graphula/Dependencies.hs
```diff
@@ -24,18 +24,29 @@ module Graphula.Dependencies
 
     -- * Non-serial keys
   , KeySourceType (..)
+  , KeySourceTypeM
+  , KeyForInsert
+  , KeyRequirementForInsert
+  , InsertWithPossiblyRequiredKey (..)
+  , Required (..)
+  , Optional (..)
```
pbrisbin, line 32: Do users bump into this stuff at all, or is it all encapsulated?

pbrisbin (approved):
The type family stuff is a bit over my head, but I'm glad to see this solved.

I think the main open question causing #70 to stall was if introducing `insertKeyed` could be done without a major version bump. Seems like you're saying it cannot, but let's do it anyway. I tend to agree.

src/Graphula/Dependencies.hs
chris-martin: The only concern is it might show up in type errors, not sure. Otherwise no.

PR #76, opened 2024-03-01
add role annotations

chris-martin:
to satisfy `-Weverything` on GHC 9.8

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -34,9 +34,6 @@ when:
 dependencies:
   - base < 5
 
-default-extensions:
-  - RoleAnnotations
-
 library:
   source-dirs:
     - src
--- a/package.yaml
+++ b/package.yaml
@@ -19,6 +19,9 @@ ghc-options:
   - -Wno-unsafe
 
 when:
+  - condition: impl(ghc >= 9.8)
+    ghc-options:
+      - -Wno-missing-role-annotations
   - condition: impl(ghc >= 9.2)
     ghc-options:
       - -Wno-missing-kind-signatures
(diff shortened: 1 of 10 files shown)


src/Graphula.hs
```diff
@@ -191,12 +191,18 @@ data Args backend n m = Args
   , gen :: IORef QCGen
   }
 
+type role Args representational representational representational
```
pbrisbin, line 194: This seems like a ton of overhead with no value. Can we `-Wno-missing-role-annotations` instead? We've done that many times for new warnings, e.g. with `missing-kind-signatures`, etc.

When it comes to our warnings choices and use of `everything`, the intent is not "seek to satisfy `everything`, only opting out what you must", it's more "`everything` + opt-out is better than opt-in" -- so there's no stigma to opting out of stuff. Would be good to document this somewhere I guess.

Unless there's some big benefit to this that I'm missing?

pbrisbin (changes requested):

src/Graphula.hs
pbrisbin: > make sure GHC agrees with what you thought a role was

FWIW, I have exactly 0 idea what any role of anything is at any point. What you have here is greek to me :shrug:

package.yaml
```diff
@@ -30,6 +30,9 @@ when:
   - condition: impl(ghc < 8.6)
     ghc-options:
       - -Wno-redundant-constraints
+  - condition: impl(ghc >= 9.8)
+    ghc-options:
+      - -Wno-missing-role-annotations
```
pbrisbin, line 35: Sorry to be a stickler, but the list of conditionals are well ordered new-to-old, so can you put this first instead of last?

pbrisbin (approved):

PR #88, opened 2025-04-01
chore(release): use published semantic-release plugin

pbrisbin:
This has been released on npm, so we no longer need the git reference.
The git reference doesn't work, in fact, because once released I stopped
committing `dist/` to version control.


pbrisbin:
Will fix this sort of error: https://github.com/freckle/graphula/actions/runs/14196998809/job/39774407636

chris-martin (approved):

Issue #44, opened 2020-06-22
Catch SomeException and decorate with seed


eborden:
Replay via `seed` is extremely useful for determining if a circumstance is flakiness or intrinsic. Non `HUnit` exceptions are not caught by `graphula` so unexpected IO exceptions cannot be replayed. Catching and decorating `SomeException` with the `seed` value can enable this workflow.

```
SqlError {sqlState = "57014", sqlExecStatus = FatalError, sqlErrorMsg = "canceling statement due to statement timeout", sqlErrorDetail = "", sqlErrorHint = ""}
```

chris-martin:
The simplest thing doesn't seem excessively offensive to me.

Maybe the display details ought to be up to context, e.g. in a spec hook. If instead of a `Text` annotation you used a type for this like

```hs
newtype GraphulaSeed = GraphulaSeed Int
```

```
uncaught exception: AnnotatedException
AnnotatedException {annotations = [Annotation @GraphulaSeed -6194587210816537995,Annotation @CallStack [("checkpoint",SrcLoc {srcLocPackage = "graphula-2.0.0.0-6RxeicdAygU6UOn5C6N1Qn", srcLocModule = "Graphula", srcLocFile = "src/Graphula.hs", srcLocStartLine = 275, srcLocStartCol = 34, srcLocEndLine = 275, srcLocEndCol = 44})]], exception = SqlError {sqlState = "57014", sqlExecStatus = FatalError, sqlErrorMsg = "canceling statement due to statement timeout", sqlErrorDetail = "", sqlErrorHint = ""}}
```

... then a spec could catch and display however it liked?

pbrisbin:
Update: I'm discarding Option 3 (second variation) and Option 5 because they both change the underlying exception type. That would cause any type-directed `catch`es to no longer catch. In my experience, there are enough "unknown unknown" situations like that in this area that it's worth avoiding.

The other options still technically change the exception type (from `e` to `AnnotatedException e`), and that has caused us a lot of pain previously for the same reason, but we accept the trade-off in other areas for this specifically; the fact that the underlying `e` is kept as-is and there is a variant of `catch` that works on either `e` or `AnnotatedException e` makes that particular type change more tolerable.

So I'm going to implement Option 1.

PR #107, opened 2026-06-16
Decorate exception-based failures with seed

pbrisbin:
### fix!: include seed on failures do to exceptions

3a67a3a6dbf668a16dbfbe693f325a6b8f3c290a

By making any `SomeException` an `HUnitFailure`, we can use our
`logFailingSeed` function to ensure we include the seed on those
failures too.

This does break exception handling happening outside of `runGraphulaT`,
since any external catches for not-`HUnitFailure` will no longer catch.

### chore: refactor FailureReason message building

a4b26d4a4b27c8bcd286fdee9edff93688b30064

Prefixing the text after using `formatFailureReason` erases the fact
that the values was a `Reason` vs `ExpectedButGot`. The new approach
prepends the text without changing the constructor.

I doubt this matters, but it was easy enough and stops disrupting anyone
inspecting such values (perhaps our JUnit formatter?).

### chore: update tests for new exceptions

73595f769ac3307500db4d2b003cc635df52490c

### chore: prefix with line breaks

b2343d534be238d5a3ca9560cca720400b9784e1

### DIFF
--- a/graphula.cabal
+++ b/graphula.cabal
@@ -1,6 +1,6 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.38.1.
+-- This file has been generated from package.yaml by hpack version 0.39.1.
 --
 -- see: https://github.com/sol/hpack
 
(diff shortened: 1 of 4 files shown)


pbrisbin:
I don't like this approach. I commented about why in #44.

PR #108, opened 2026-06-16
Annotate SomeException with Graphula seed

### LINKED ISSUE #44: Catch SomeException and decorate with seed

pbrisbin:
**Open Questions**:

1. Is this feature worth changing the exception type thrown?
2. Is that change in exception type a major version bump?

You can see in my second commit the kind of thing our users might have to do. It's an unusual thing to be doing, but it will be broken by this "fix".

---

### fix: annotate SomeException with Graphula seed

ad00327f2f1c604548d25f49fe795fb1ba61ce97

The seed is useful to re-run in a deterministic way, which is why we
prepend it to any `HUnitFailure` exception messages. This commit adds
the same treatment to all other exceptions by catching `SomeException`
and using `checkpoint` to add it as an annotation.

Here is an example:

```
Before:

  uncaught exception: GenerationFailureMaxAttemptsToInsert
  GenerationFailureMaxAttemptsToInsert (Just "entity already exists by this key") School}

After:

  uncaught exception: AnnotatedException
  AnnotatedException {annotations = [Annotation @GraphulaSeed (GraphulaSeed (-8068470525722554834)),Annotation @CallStack [("checkpoint",SrcLoc {srcLocPackage = "graphula-2.0.0.0-ArtCvOVJMvo1gDWVYi6qKp", srcLocModule = "Graphula", srcLocFile = "src/Graphula.hs", srcLocStartLine = 272, srcLocStartCol = 8, srcLocEndLine = 272, srcLocEndCol = 18})]], exception = GenerationFailureMaxAttemptsToInsert (Just "entity already exists by this key") School}
```

Changing the thrown exception type is a risk, since any outer `catch`es
our users (or other libraries) have in place may no longer catch it. But
using `checkpoint` and changing `e` to `AnnotatedException e` should be
tolerable, due to the fact that `AnnotatedException.catch` exists, which
catches both types at once.

That is why the output is relatively ugly, because we want to rely on
`checkpoint` and `AnnotatedException`, so this is what you get.

I also noticed that the existing code changes the `HUnitFailure`
constructor used (both `Reason` and `ExpectedBotGot` are turned into
`Reason` by way of `formatFailureReason`). Ideally, we would not do that
(and there's no real need), but I'm leaving it alone for now, erring on
the side of a smaller diff for this change.

Closes #44.

### chore: update README to use annotated-exception's try

ad5f5b7f8e4d1893aed4a18d465344f9cf751302

This is necessary because now the example throws an
`AnnotatedException GenerationFailure` and not a bare
`GenerationFailure` exception.

This is an example of the sort of change our users might have to make
because of the seed annotation feature.


src/Graphula.hs
```diff
@@ -262,8 +263,13 @@ runGraphulaT mSeed runDB action = do
...
 
-logFailingSeed :: MonadIO m => Int -> HUnitFailure -> m a
-logFailingSeed seed = rethrowHUnitWith ("Graphula with seed: " ++ show seed)
+newtype GraphulaSeed = GraphulaSeed Int
+  deriving stock (Show)
+
+logFailingSeed :: MonadUnliftIO m => Int -> SomeException -> m a
+logFailingSeed seed ex = case fromException ex of
+  Just hf -> rethrowHUnitWith ("Graphula with seed: " <> show seed) hf
+  _ -> checkpoint (Annotation $ GraphulaSeed seed) $ throwIO ex
```
cdmren, line 272: At this point I think it might make more sense to use the [built-in exception context support](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Exception.html#v:addExceptionContext) from base (since base 4.20, so it's available in a stackage LTS) instead of annotated-exception.



```suggestion
  _ ->
  throwIO $
    addExceptionContext (GraphulaSeed seed) $ -- guard this line in a CPP conditional
    ex
```

pbrisbin: Well that certainly sounds much better, and it doesn't change the exception type (I assume), so the downsides are gone. And with the CPP, we don't have to drop any support.

pbrisbin:
I'm going to take a new swing in a new PR.

PR #113, opened 2026-07-31
feat: add seed via exception context

### LINKED ISSUE #44: Catch SomeException and decorate with seed

pbrisbin:
If using `base >= 4.20`, we now add the graphula seed as an annotation in the exception context[^1][^2].

This will be present for all exceptions (closes #44), but it's not readily visible anywhere. It must be extracted specifically via hooks (see the tests). Therefore, this hasn't replaced the textual-addition to `HUnitFailure`, it is in addition to.

[^1]: https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Exception-Context.html
[^2]: https://well-typed.com/blog/2026/05/lay-annotation-land/


cdmren:
I've been playing with removing annotated-exception in the curricula project so I understand a bit more what exception context actually is, now, and I guess it's not quite what I thought it was when I suggested doing that here.

There's an option where we instruct users to add something to a top-level SpecHook that would catch graphula's exception information (be that a seed in an exception context attached to an hunit exception, or a custom exception type containing the seed as a field) and re-throw as an hunit exception with information in the failure reason string. There's an argument to be made that making the hook responsible for error formatting is the only reasonable choice hspec gives you.

I'm not sure it's worth it though.

> should we keep this behavior of stuffing it into strings

Seems right.

But is there any harm in *also* adding it as exception context? It seems like that would allow us to address #44 (make the seed programmatically obtainable from any type of exception) without changing either the exception type or the hunit failure reason string.

pbrisbin:
> But is there any harm in also adding it as exception context?

Makes sense. I'll try that.

pbrisbin:
OK, I've done it that way. I clobbered what was here and did a "make the change easy / make the easy change" method in two commits:

### [chore: refactor throw vs exception modification](https://github.com/freckle/graphula/pull/113/commits/a04535cc82065d26de66b9b73a373c992ea846f5)
[a04535c](https://github.com/freckle/graphula/pull/113/commits/a04535cc82065d26de66b9b73a373c992ea846f5)

This moves the `throwIO` out of the function that is concretely
`HUnitFailure`. This is because I'll next be adding something that needs
to occur between those two points.

### [feat: add seed to exception context, if supported](https://github.com/freckle/graphula/pull/113/commits/21dfb5d3c1e4e88eeaf75610e324756674451f67)
[21dfb5d](https://github.com/freckle/graphula/pull/113/commits/21dfb5d3c1e4e88eeaf75610e324756674451f67)

If using `base > 4.20` we add the graphula seed to the exception
context of any thrown exceptions.

Closes https://github.com/freckle/graphula/issues/44 as best we can. The information won't appear anywhere
automatically, but can be pulled out (e.g.) via `SpecHook` for those
that know (see README). This is a compromise. Give how `hspec` works
today, any other method that would make the seed more visible for
non-expectation-failure exceptions comes with too many negative
trade-offs.

src/Graphula.hs
```diff
@@ -133,6 +134,9 @@ module Graphula
   , NodeOptions
   , GenerateKey
   , NoConstraint
+#if MIN_VERSION_base(4,20,0)
+  , GraphulaSeed(..)
```
cdmren: The comment above deprecates this section, did you mean to add this here?

src/Graphula.hs
```diff
@@ -133,6 +134,9 @@ module Graphula
   , NodeOptions
   , GenerateKey
   , NoConstraint
+#if MIN_VERSION_base(4,20,0)
```
cdmren: I wonder if the CPP guard here causes more confusion than it alleviates; `GraphulaSeed` could be defined and exported regardless of the `base` version.

cdmren (dismissed):
Might be worth adding a test, I find Typeable stuff very finicky.

src/Graphula.hs
pbrisbin: I just didn't see that part of the comment. I only saw "Lower-level". I didn't realize we had deprecated re-exports here.

src/Graphula.hs
pbrisbin: Hmm, yeah. There's also the fact that `runGraphulaT` accepts the seed as an `Int`. This type existing would certainly make me expect it to take a `GraphulaSeed`.

I wonder if `GraphulaExceptionContext { seed :: Int }` is better.

* Naming implies it's only used when "exception context" is a thing (I'll also document that)
* Wouldn't be expected to be the `runGraphulaT` argument
* May make sense in a `Graphula.ExceptionContext` module, to isolate `CPP`

pbrisbin:
I'll think about a test when I touch this again, but IMO there's not a lot going on. I'm calling a simple function from `base` and expecting it to do what it says on the tin :shrug:

cdmren (changes requested):
See #115

cdmren (approved):

pbrisbin:
We need a third reviewer now because Chris is last pusher and it's my PR.

mjgpy3 (approved):

